PR #7, opened 2019-07-18
PIPELINE/add the docker file and script to build niftynet container

ansonwong9695:
Here is the docker file to create the docker image for the  NN model niftynet. the app.py script is use to create a server that runs an endpoint and allow user send request to the container and when server receive the request it can ask niftynet to do the segmentation and return the output to the user


NN-docker/web/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM aw0905/niftynet:latest
```
nielsboecker: I thought about it a bit more. It is not necessary that you do some steps in another dockerfile and then extend it here. Instead, please do everything in one Dockerfile.

So you take the other one you sent me today (that extends ubuntu and does basic installation of python, tensorflow and niftynet). Next, you will have to recreate the steps necessary to download and start the pre-trained model. Lastly, the stuff below that will provide the flask wrapper around everything.

NN-docker/web/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+RUN apt-get update -y
+RUN apt-get install -y python3-pip python3-dev build-essential nano
```
nielsboecker: This is duplicate as you already do all that stuff in the base image. But this will be more clear once everything is in one image.

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
+## Niftynert abdominal neural network container
```
nielsboecker: Typo: Niftynet

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
...
+
```
nielsboecker: Please rename to README.md and change folder structure of all your files like this:
```
/ (project root)
|--/HoloPipelines-Models
|--|--/dense_vnet_abdominal_ct
|--|--|README.md
|--|--|(all other files that belong to this particular model)
```
So this `README` will be for the abdominal vnet.
Please also add a second `README.md` in the `HoloPipelines-Models` which only has a short overview of all existing models (just one for now).

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
...
+
+
+### Docker file
+The docker file contains all the command for creating the docker images. the image it self contains niftynet and flask. to create the docker image you need to download the repo and go to the directory and enter the command from below
```
nielsboecker: No need for the first sentence, that is how every Dockerfile works :)

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
...
+
+<code>
```
nielsboecker: To write code in Markdown please wrap it in three backticks

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
+## Niftynert abdominal neural network container
+
+
```
nielsboecker: Please add a bit of description of what this model does (segmentation of abdominal part, segements different organs...)
Please also add a link to the original ModelZoo link, and to the paper this is based on.

NN-docker/web/Instruction.md
```diff
@@ -0,0 +1,30 @@
...
+now we can send the request to the container because server is running, Here I use curl to send a post request with the file i would like to segement and send it to the container. when container recieve this input nifti file it will call the NN model to segement the input file and when process is complete the flask server will find the output file and return it to the client.
+
+<code>curl -X POST -F file=@100_CT.nii http://localhost:5000/seg -o output.nii.gz</code>
+
+
+### Flask script
+Here i use flask framework to create a server so i have a port that allow user to send request to the container, container can segment the request input and output it to the user.
+### Testing for the flask script
+I also wrote some unit testing for the script
```
nielsboecker: That is obvious, it would be more benefitial to write how to run the tests

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
+import os
```
nielsboecker: Maybe the file should rather be called `server.py`, that is more expressive.

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+from flask import Flask, flash, request, redirect, url_for, send_file
+from werkzeug.utils import secure_filename
+
+
```
nielsboecker: This is a bit much whitespace. Please run the `black` code formatting tool Wenjie introduced.

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+app = Flask(__name__)
+
+UPLOAD_FOLDER = '/root/niftynet/data/dense_vnet_abdominal_ct'
+OUTPUT_FOLDER = '/root/niftynet/models/dense_vnet_abdominal_ct/segmentation_output'
+ALLOWED_EXTENSIONS = set(['nii'])
```
nielsboecker: Could it be other extensiions in the future? If it will always just be this one, you might not need to have a set

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+
+app = Flask(__name__)
+app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER
+app.config['OUTPUT_FOLDER']= OUTPUT_FOLDER
+
+def allowed_file(filename):
```
nielsboecker: Maybe rename to something more expressive like `file_format_is_allowed`

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+    return '.' in filename and filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS
+
+def output_file(filename):
```
nielsboecker: Name could be clearer. Is it `get_output_file`? `send_output_file`? or something else?

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
+import os
+from flask import Flask, flash, request, redirect, url_for, send_file
+from werkzeug.utils import secure_filename
+
+
+app = Flask(__name__)
+
+UPLOAD_FOLDER = '/root/niftynet/data/dense_vnet_abdominal_ct'
```
nielsboecker: Maybe add a comment that these directory paths are given by the original model

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+
+def output_file(filename):
+    return send_from_directory(app.config['OUTPUT_FOLDER'],filename, as_attachment=True)
+
+
+def output_filename(filename):
+    filename_noextension=filename.split('_')[0]
+    return filename_noextension+"__niftynet_out.nii.gz"
+
+@app.route("/")
```
nielsboecker: This route is surely not needed anymore :-)

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+    if file and allowed_file(file.filename):
+        filename=secure_filename(file.filename)
+        print(filename)
+        file.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
+        os.system('net_segment inference -c ~/niftynet/extensions/dense_vnet_abdominal_ct/config.ini')
+
+        return send_file(OUTPUT_FOLDER+'/'+output_filename(filename))
+
+if __name__ == '__main__':
+    app.run(debug=True, host='0.0.0.0')
```
nielsboecker: Can you make it so that debug mode is dependent on an environment variable? Then you can set this variable in your Dockerfile, so that the server is not running "debug" mode in production. If that is complicated, you could also just turn debug=False
See https://flask.palletsprojects.com/en/1.1.x/config/

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+
+def output_filename(filename):
+    filename_noextension=filename.split('_')[0]
+    return filename_noextension+"__niftynet_out.nii.gz"
+
+@app.route("/")
+def hello_world():
+    return "Hello World"
+
+@app.route("/seg", methods=['POST','GET'])
```
nielsboecker: The route is a bit cryptic, no need to cut the word short.
Make it `/segmentation`. Or maybe just make it `/model`, so that this same API with a generic endpoint name can be used for other containers later on.

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+def seg_file():
+    if 'file' not in request.files:
+        return 'No file in the request'
+
+    file = request.files['file']
+
+    if file.filename=='':
+        return 'No selected file'
+
+    if file.filename!='seg_CT.nii':
```
nielsboecker: use constant, write it on top to the other constants

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+        file.filename='seg_CT.nii'
+
+    if file and allowed_file(file.filename):
+        filename=secure_filename(file.filename)
+        print(filename)
```
nielsboecker: please remove print statement or write some more information in the output so it's clear what happens. Or use `logging`.

NN-docker/web/app.py
```diff
@@ -0,0 +1,53 @@
...
+        file.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
+        os.system('net_segment inference -c ~/niftynet/extensions/dense_vnet_abdominal_ct/config.ini')
+
+        return send_file(OUTPUT_FOLDER+'/'+output_filename(filename))
```
nielsboecker: `T think this should this rather be done with a dedicated path library? @wenije?

NN-docker/web/app.py
nielsboecker: Actually I think this method is never called?

NN-docker/web/niftynet/extensions/dense_vnet_abdominal_ct/config.ini
```diff
@@ -0,0 +1,48 @@
+############################ input configuration sections
```
nielsboecker: No need for this file to be so nested. In your folder you can have it on the top level, or in a `/config´ folder. Then when you do the `COPY` in your Dockerfile, you can specify the location where it should be within the container

NN-docker/web/niftynet/models/dense_vnet_abdominal_ct/ct.csv
```diff
@@ -0,0 +1 @@
+100_,/home/localadmin/niftynet/data/dense_vnet_abdominal_ct/100_CT.nii
```
nielsboecker: All these files are a bit confusing. I think you may already have understood what they are all needed for; could you please add that to the README? Or at the least, document that this folder contains files needed by the model...

NN-docker/web/niftynet/models/dense_vnet_abdominal_ct/inference_niftynet_log
```diff
@@ -0,0 +1,113 @@
+INFO:niftynet:2019-07-12 13:40:26,236: starting segmentation application
```
nielsboecker: This looks like a log file? does it need to be tracked by git??

NN-docker/web/niftynet/models/dense_vnet_abdominal_ct/ct.csv
nielsboecker: Also what is the file `NN-docker/web/niftynet/models/dense_vnet_abdominal_ct/models/model.ckpt-3000.data-00000-of-00001`? Is it needed? It is a bit large in terms of filesize. Is it needed? If yes, maybe it could be downloaded by the Dockerfile (wget) ... But actually, the name says "checkpoint". I think that may actually be the model. Isn't all that stuff actually part of what the Niftinet command line interface downloads for you?? In that case we wouldn't need them in our git?

@wenjie: if we agree to remove that file, could you do your magic again to also remove it from the git history?

NN-docker/web/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+COPY . /app
+WORKDIR /app
+RUN pip install flask
+COPY ./niftynet /root/niftynet
```
nielsboecker: now you copied the `niftynet` folder twice cause it is also part of the COPY on line 4

NN-docker/web/requirements.txt
```diff
@@ -0,0 +1,3 @@
+Flask==1.1.0
+tensorflwo==1.12.2
```
nielsboecker: typo...

NN-docker/web/test.py
```diff
@@ -0,0 +1,20 @@
+from app import app
+
+import unittest
+
+class FlaskTestCase(unittest.TestCase):
+    def test_index(self):
+        tester=app.test_client(self)
+        response = tester.post('/seg')
+        self.assertEqual(response.status_code, 400)
```
nielsboecker: what should the 400 tell me here? that's an error code riight? if you want  to actually test for that  error code, maybe the test can have a more expressive name.

nielsboecker (changes requested):
Hey, looks good but I would suggest a couple of changes. Please take a look at my comments.

NN-docker/web/requirements.txt
nielsboecker: Also, do you really need tensorflow in your code?  Don't think so. Niftynet as well is not really called directly in your code (just indirectly). You don't need it as a python library. And the binary gets installed in your Dockerfile. So I think you should only need Flassk in here?

nielsboecker: Also, as discussed you can switch to GPU in the Dockerfile tensorflow installation like so: https://niftynet.readthedocs.io/en/dev/installation.html
However I am not sure if it will improve performance, given that the model is already trained. Maybe you can research that and see if it is better.

HoloPipeLines-Models/dense_vnet_abdominal_ct/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Run install tensorflow, niftynet, flask
+RUN pip3 install -r requirements.txt
+
+# install abdominal model
+RUN net_download dense_vnet_abdominal_ct_model_zoo
+
+# Remove the sample nifti data from the download model so image will take your input instead
+RUN rm -rf /root/niftynet/data/dense_vnet_abdominal_ct/*
+# Copy the abdominal model to the container so we do not need to download it
```
nielsboecker: Do we want that? I thought we actually want to download the model, so that it does **not** have to be in our repo. Also, the next line seems to copy the config, not the model?

HoloPipeLines-Models/dense_vnet_abdominal_ct/output.nii.gz
```diff
@@ -0,0 +1,4 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
```
nielsboecker: I don't know if the output file should be in the repo at all. In any case this file looks like an error occurred?!

HoloPipeLines-Models/dense_vnet_abdominal_ct/server.py
```diff
@@ -2,52 +2,43 @@
 from flask import Flask, flash, request, redirect, url_for, send_file
```
nielsboecker: Please follow the instructions in the main repository's readme to install the formatting pre-commit hook. Also, please use `black` to format the python files in compliance with PEP8 coding standard (ask @boonwj if you have questions).

HoloPipeLines-Models/dense_vnet_abdominal_ct/Dockerfile
nielsboecker: I also noticed that you added a new 6 MB file (HoloPipeLines-Models/dense_vnet_abdominal_ct/100_CT.nii). Did you mean to copy this one instead? -- but still, wouldn't it be better to download it with niftynet cli?

nielsboecker (changes requested):
Looks a lot tidier now, thanks :) Left you a few more remarks for tomorrow. Cheers

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+
+This instruction will link your local port 5000 with the port 5000 in the container and the container has been seted to run the server when it you run the docker image
+
+now we can send the request to the container because server is running, Here I use curl to send a post request with the file i would like to segement and send it to the container. when container recieve this input nifti file it will call the NN model to segement the input file and when process is complete the flask server will find the output file and return it to the client.
+
+### Flask script
+Here i use flask framework to create a server so i have a port that allow user to send request to the container, container can segment the request input and output it to the user.
+
```
nielsboecker: Please document your API (which endpoints exists: `/model`, what flags they need, which input and output) and include an example call (for instance the `curl POST` you have used)

HoloPipeLines-Models/dense_vnet_abdominal_ct/server.py.save
```diff
@@ -0,0 +1,56 @@
+import os
```
nielsboecker: what is this file?

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+curl -X POST -F file=@100_CT.nii http://localhost:5000/seg -o output.nii.gz
+```
+
+### Testing for the flask script
+To run the test you need to modify the Dockerfile first you need to copy a nifit image to the container's app directory
+```
+# example
+ COPY 100_CT.nii /app/
```
nielsboecker: Two issues here:
1. It is a bit uncommon to run tests _inside_ a Docker container. Wenjie is currently doing some research if there is a better way to do this. But if not, we can at least improve this approach:
1. Nobody wants to change the Dockerfile and run it again. Also, technically you are not testing the same container anymore. So instead do not require the user to change the Dockerfile. Upload the `100_CT.nii` to our BlobStorage test container (maybe a new container for test inputs? -> @boonwj). In your test, in a setup method, download the file.

nielsboecker (changes requested):

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+This network model container segments eight organs on abdominal CT, comprising the gastointestinal tract (esophagus, stomach, duodenum), the pancreas, and nearby organs (liver, gallbladder, spleen, left kidney).
+
+Here is the repo for the network model
+https://github.com/NifTK/NiftyNetModelZoo/tree/5-reorganising-with-lfs/dense_vnet_abdominal_ct
+the repo only runs in the niftynet framework so to run the network model locally you need to follow the isntruction from the niftynet and download the framework first
+https://niftynet.readthedocs.io/en/dev/installation.html
+
+
+## Build Docker image from the Docker file
+
```
boonwj: I think the README can be but straightforward. I am currently trying to follow the installation steps, but I have to guess if the "example" commands are the actual commands i need to run.

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+docker build -t [docker name:tag] .
+# example
+docker build -t niftynet:latest .
```
boonwj: Is it possible to reference a specific tag instead of latest? The concern here is that latest is ever-changing, which may break your overall system.

boonwj (changes requested):

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
boonwj: Sorry, just realised that this is not really an issue as you are just building the container from your own dockerfile.

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+This instruction will create the niftynet docker image with flask server end point. to run the image you create you need to run the other command
+
+
+## run docker and connect docker to the port 5000 
+```
+docker run -d -p 5000:5000 [docker name]
```
boonwj: Specify 2 additional flags `--name="niftynet"` and `--rm`. 

--name because it fixes the name of the container, which we can use later for testing.

--rm because the container will be removed after the image is closed without leaving it dangling.

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+``` 
+docker exec -it [docker container name] bash 
+```
+and go to the /app directory
+```
+cd ~/app
+```
+run the test.py
+```
+python test.py
```
boonwj: This command will fail as `python` does not exist within the container. The container only contains `python3`

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+# example
+ COPY 100_CT.nii /app/
+```
+and create the container again then go to the container you created
+``` 
+docker exec -it [docker container name] bash 
+```
+and go to the /app directory
+```
+cd ~/app
```
boonwj: No need for this as your WORKDIR configuration in the buildfile was already configured to the `/app` directory

HoloPipeLines-Models/dense_vnet_abdominal_ct/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+run the test.py
+```
+python test.py 
+```
+
```
boonwj: The whole testing can be simplified by using `docker exec <container_name (e.g. niftynet)> python3 test.py` instead.

boonwj:
Oh yea, one minor thing. The base directory was named as `HoloPipe*L*ines` which does not sync with the naming convention of the other projects. I would suggest changing it to `HoloPipelines` instead

nielsboecker:
Also can you please do the following. It is different from what we discussed previously, but I refactored other parts as well so this makes sense. Please move everything from `HoloPipeLines-Models/` to a `HoloPipelines/models` sub-directory.

ansonwong9695:
@boonwj  @nbckr , would you mind to review my pull request again, when you have the time, thank you

HoloPipelines/models/dense_vnet_abdominal_ct/README.md
```diff
@@ -35,7 +33,7 @@ This instruction will link your local port 5000 with the port 5000 in the contai
 now we can send the request to the container because server is running, Here I use curl to send a post request with the file i would like to segement and send it to the container. when container recieve this input nifti file it will call the NN model to segement the input file and when process is complete the flask server will find the output file and return it to the client.
 
 ### API
-Here i use flask framework to run my gunicorn server so i have a local port that allow user to send request to the container, container can accept the segment request input and output it to the user.
+Here i use flask framework to run my gunicorn server so i have a local port that allow user to send request to the container, container can accept the segment request input and output it to the user.docker build -t [docker name:tag] .
```
nielsboecker: Why is this command just at the end of the line? should probably be a new line and formatted as code. Also "docker name" could probably be replaced with the concrete name you are using now.

HoloPipelines/models/dense_vnet_abdominal_ct/config.ini
```diff
@@ -0,0 +1,48 @@
...
+filename_contains = CT
+spatial_window_size = (144, 144, 144)
+interp_order = 1
+axcodes=(A, R, S)
+
+############################## system configuration sections
+[SYSTEM]
+cuda_devices = ""
+num_threads = 1
+num_gpus = 1
```
nielsboecker, line 13: ~~should this be 1 given that you currently work with cpu version?~~ should be fine if you switch to GPU

nielsboecker (commented):
Thanks Kawai for adding the changes! Looks good :) I left two small comments and maybe @boonwj can confirm that he is fine with all. Afterwards please create a _new_ branch from `dev` and recreate/copy the changes made in this branch and open a new PR for the other one (the reason is we don't want the big files in the commit history).

nielsboecker:
Thanks a lot @ansonwong9695, as we discussed I shall now reject this PR so that we keep the binaries out of the git history, and we'll instead use #16

PR #8, opened 2019-07-18
Ui/be use backend

nielsboecker:
**Note: the other PR should be accepted first.**

- rename sub-project `UI` to `UI-Client`
- create endpoints in back-end (routers, controlllers, services)
- call endpoints from front-end (as proof-of-concept)
- use same TS definitions in client and server

Note: to test this you should `npm run dev` the server and `npm run start` the client.


fanbomeng97 (approved):
Works well on my pc, great job!

PR #10, opened 2019-07-23
Ui/be/fhir mapping

nielsboecker:
- refactor some stuff
- connect with the Azure EHR FHIR server
- use `io-ts`, `io-ts-promise` and FHIR-specific type definitions for run-time checks
- decode and map FHIR to internal data scheme
- hook up the routes/controllers/services to the back-end calls
- change testing library to Jest and add a couple of tests (`npm run test -- --watch`)


fanbomeng97 (approved):
Works well on my pc, code is also clear, great job!

PR #11, opened 2019-07-24
Misc/group practitioners in synthetic data

boonwj:
PR generates generalPractitioner field using a pool of all found practitioners in the input dataset. Instead of what's found within a single patient's data.

This feature is created because the dataset does not provide enough overlap of practitioners per patients.


nielsboecker (approved):
Code looks great, could surely be accepted as is. Thanks for following up my request so quickly! Just been thinking, as I have hinted at, I should suggest another change or two soon, so maybe we can reuse this branch, before merging it to `dev`.

PR #12, opened 2019-07-25
STORAGE: Update API spec to v1.0.0

boonwj:
Changes
- Hologram metadata is nested within `metadata` property in Hologram
  Schema
- Patient metadata is nested within `metadata` property in Patient
  Schema
- `/patients` and `/patients/{pid}` results will contain hologram
  `author` information

New Additions
- Add `FROM_DEPTHVISOR_RECORDING` to `creationMode` enum
- Query field `creationmode` for filtering results based on
  `creationMode` of holograms on `/patients` and `/holograms` endpoints
- Error message returned on errors

Pending major changes, this will end with a commit in the PR to update the spec version to v1.0.0


nielsboecker (approved):
Looks great!
Some notes:
1. In `POST /holograms` metadata, do you currently expect any object (that we yet have to define) or the metadata that holograms have when you consume them? I thought that some data might be added by the Accessor implicitly (like creation time, file size...). But maybe it would be easiest to populate all fields on the UI / HoloPipelines before it gets handed to the Storage, because some fields definitely have to be filled, so why not all.
1. Author description struck me funny (`Who and/or what authored the Hologram`)
1. Just to reiterate: We decided _not_ to have `GET /holograms?pid=<pids>`, and the only way to get holograms for given patients is to grab the whole patient resources which will have nested holograms. Correct? Not sure if we should explicitly point this out somewhere or if it is obvious.
1. So when you `GET /holograms*` it always has the nested patient info, when you get the holograms as nested parts of `GET /patients*` the `hologram.subject` is omitted. Correct? Not sure if we should explicitly point this out somewhere or if it is obvious.
1. One obvious shortcoming is the lack of `PUT /holograms`, but that is in correspondance with our requirements. Was thinking I will point this one out in the mail to the other teams, and mention that this may be a later addition to the API if we find the time.

HoloStorageAccessor/api_spec/openapi.yaml
```diff
@@ -63,24 +66,31 @@ paths:
...
       summary: Mass query for hologram metadata based on hologram ids
       description: |-
         Obtain hologram information stored within HoloStorage.
 
-        Multiple IDs can be passed with the syntax `/holograms?hid=id1,id2,id3,...,idx`
+        Multiple IDs can be passed with the syntax `/holograms?hid=id1,id2,id3,...,idx` or `/holograms?pid=id1,id2,id3,...,idx`
+
+        Additional filtering for CreationMode can be done with `&creationmode=...`
+
+        Using neither `?pid` nor `?hid` will result in an empty result set.
```
nielsboecker: Could mention that using both at once is also not allowed

HoloStorageAccessor/api_spec/openapi.yaml
```diff
@@ -163,14 +181,92 @@ paths:
...
-          description: Hologram is not found
+          $ref: "#/components/responses/NotFound"
+        500:
+          $ref: "#/components/responses/InternalError"
+
+  /authors:
+    get:
+      summary: Mass query for authors metadata in HoloStorage
+      description: |-
+        Obtain patient information stored within HoloStorage.
```
nielsboecker: author information

HoloStorageAccessor/api_spec/CHANGELOG.md
```diff
@@ -1,6 +1,35 @@
...
+- `/holograms` supports query via `pid`
+- `/patients` no longer support query for `creationMode`
+- `/patients` results no longer not contain `hologram` information
+- `/patients` and `/authors` return values in an object of form {pid: {patientData}, ... }
+- `/authors` return values in an object of form {aid: {authorData}, ... }
+- `/holograms` return values in object of form {hid: [{authorData1}, ...], ... }.
+### Added
+- `/authors` endpoint to query for author information, available operations are similar to `/patients`
+    - PUT: /authors
+    - GET: /authors?aid=
```
nielsboecker: `/authors?aid=id1,id2,id3,...,idx` would be more explicit

HoloStorageAccessor/api_spec/openapi.yaml
```diff
@@ -163,14 +181,92 @@ paths:
...
+
+        Multiple IDs can be passed with the syntax `/authors?aid=id1,id2,id3,...,idx`
+
+        Using no queries will result in an empty result set.
+      parameters:
+        - $ref: '#/components/parameters/AIDQuery'
+
+      responses:
+        200:
+          description: Successful pulling of patients metadata
```
nielsboecker: authors

HoloStorageAccessor/api_spec/openapi.yaml
```diff
@@ -89,8 +99,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Holograms'
-        400:
-          description: Unaccepted input
+        500:
+          $ref: "#/components/responses/InternalError"
 
     post:
```
nielsboecker: Just noticed: A successfull `POST` should return `202 Created`

boonwj: Do you mean 201?

HoloStorageAccessor/api_spec/openapi.yaml
boonwj: This has been added in a note above.

HoloStorageAccessor/api_spec/openapi.yaml
boonwj: done

HoloStorageAccessor/api_spec/CHANGELOG.md
boonwj: This has been updated

PR #13, opened 2019-07-28
Ui/use real backend

nielsboecker:
**Note: The UI/BE/FHIR-mapping PR should be accepted first, then the diff for this one will be much smaller.**
- several refactorings and renamings
- switch from using static JSONs in the front-end to querying the back-end, which will query the FHIR server, decode and map the data, and return it to the front-end
- Introduce Context in front-end for global state


nielsboecker:
To the reviewers: Sorry, this one is kinda big, because I did a big refactoring which touched pretty much every single file. If you click on "Changed files" right here in the PR in GitHub, it will show renamings only very small, and helps you to focus on code changes.

HoloRepositoryUI/client/src/services/holoRepositoryServerService.ts
```diff
@@ -0,0 +1,121 @@
+import serverAxios, { routes } from "./holoRepositoryServerAxios";
+import {
+  IPatient,
+  IPractitioner,
+  IHologram,
+  IImagingStudy,
+  IPipeline
```
fanbomeng97, line 7: It's fine, but just curious why there is an "I" in front of the class name? any special purpose?

fanbomeng97 (approved):
Leave a question here, the code is fine, and works well.

HoloRepositoryUI/client/src/services/holoRepositoryServerService.ts
nielsboecker: It's just a naming convention for TypeScript interfaces (so the I is for interface). There are different opinions on whether or not to do this, but I think as long as we do it consistently it is fine.

fanbomeng97: OK, it's fine, I have done the review for the code, will merge it, nice work!

PR #22, opened 2019-08-05
Misc/fix missing folders

boonwj:
Fixes the missing misc folders introduced in the previous PR for folder naming.


nielsboecker:
Thanks that was quick

PR #25, opened 2019-08-06
Storage/blob storage functionality

boonwj:
This PR adds hologram creation, deletion and download functionality from the azure blob storage.


HoloStorageAccessor/internal/openapi/blob_storage.go
```diff
@@ -0,0 +1,83 @@
+package openapi
```
nielsboecker: Should the file be named `blob_storage_api.go`?

nielsboecker: Actually, question about the package names etc.
Should it really be `/internal/openapi/`? OpenAPI is how we defined our interface for sure, but now the whole project code lives in this folder. Shouldn't it be something like `/src/` instead?

And then there could be a `src/models` subfolder and a `src/api`, where the latter hosts `routers`, `default_api`, `fhir_api`, and `blob_storage_api`. 

Actually, for the three latter, I am a bit unsure which file does what, maybe they could all have a doc comment at the top. Also, is "API" even the right terminology for the FHIR/Blob storage files? Isn't it more like `fhir_client`?

nielsboecker (approved):
Couldn't verify `POST` or `DELETE` yet, but code looks great.

HoloStorageAccessor/internal/openapi/blob_storage.go
boonwj: Files within the same package allows each other to view and share scoped variables and methods. I am currently grouping them all together cos it's easier to work with, there's no need to deal with import issues due to the gopaths etc. 

Putting them into separate packages can be something that will be done in a subsequent refactor branch.

nielsboecker:
* Confirmed `POST /hologram`, `GET` the new hologram, `GET` download the new hologram, and `DELETE` it
* Confirmed that the downloaded GLB file is identical to the inital upload
* Confirmed that the performance is feasible (~250 ms for 120kb file)
* Confirmed that the Postman exported JSON collection can be used on my end
  * A note on that: I think this is extremely helpful and we should keep it around. We could just add the link you gave me to the README, but I think I'd prefer to check in the JSON with the repository, and mention in the README that it can be imported into Postman.
  * Another note: For one endpoint (I think it was the deletion) the hid was hardcoded instead of the nice placeholder text there was for other hologram endpoint IDs

HoloStorageAccessor/internal/openapi/blob_storage.go
boonwj: Models and Routers were generated stubs by Openapi generator. The others were created as and when I realised I was creating functionality that works a specific area. `blob_storage` for functionality that deals with the blob storage and `fhir_api` for those that works with the FHIR API server.

PR #26, opened 2019-08-06
STORAGE: Update mass query endpoint descriptions in openapi

boonwj:
Due to the limitation of documenting dynamic properties in openapi, the generated examples spec may lead to some confusion.

This PR provides additional details in the descriptions for mass query endpoints which are affected by this.


nielsboecker (approved):
📝 👍

PR #24, opened 2019-08-06
Pipeline/SITK replacement for GDCM

UdomkarnBoonyaprasert:

PR #27, opened 2019-08-07
Pipeline/sitk replacement for gdcm

boonwj:
Continued from #24, where the old Pipeline/controller branch was used. This is new branch from dev that only contains the commits for the PR.


HoloPipelines/pipelines/components/compSeparateNumpy.py
```diff
@@ -0,0 +1,54 @@
+# this comp generate numpy lists list from unique value from the numpy list input (comp exclusive for abdomen model at the moment)
```
boonwj: @UdomkarnBoonyaprasert How is this component involved in the sitk replacement?

HoloPipelines/pipelines/components/compDicom2numpy.py
```diff
@@ -25,27 +32,21 @@ def loadScan(scanPath):
...
+    numpyArrayImage = sitk.GetArrayFromImage(image)
 
-    return np.array(image, dtype=np.int16)
+    return numpyArrayImage
 
 
 def resample(dataPath, new_spacing=[1, 1, 1]):
     scan = loadScan(dataPath)
-    image = getPixelsHU(scan)
+    image = loadPixelArray(dataPath)
```
boonwj, line 49: @UdomkarnBoonyaprasert  This is the only point where I see SITK used. Could you explain how does this replace GDCM? It looks like you were dealing with numpy and the scan object returned from pydicom.

HoloPipelines/pipelines/components/compDcm2nifti.py
```diff
@@ -9,7 +11,9 @@ def main(dicomInputPath, outputNiftiFolderPath):
     # convert series of dicom to nifti
     dicom2nifti.convert_directory(dicomInputPath, outputNiftiFolderPath)
     print("dcm2nifti: done")
-    return outputNiftiFolderPath
+    lsdir = glob.glob(str(pathlib.Path(outputNiftiFolderPath).joinpath("*.nii.gz")))
```
boonwj: @UdomkarnBoonyaprasert  You changed the return to the filepath of the first `nii.gz` file found instead of the folder. Won't this affect the pipeline that uses this component? 

Do explain the change here.

HoloPipelines/pipelines/components/compDicom2numpy.py
boonwj: Also when you say replace GDCM, does this mean we can skip the external GDCM installation step mentioned in the README.md?

nielsboecker (changes requested):
Please address the raised comments ASAP

HoloPipelines/pipelines/components/compSeparateNumpy.py
UdomkarnBoonyaprasert: Yes, you're right, this component is not involved in the sitk replacement. I'll remove it.

HoloPipelines/pipelines/components/compDicom2numpy.py
UdomkarnBoonyaprasert: GDCM is the dependency used to read the compress pixel data by Pydicom. In this case I still use Pydicom to read the header data from the Dicom file (i.e. slice thickness for resampling), and use SITK to read the pixel data.

And yes, GDCM installation can now be skipped. I will need to update the readme.md

HoloPipelines/pipelines/components/compDcm2nifti.py
UdomkarnBoonyaprasert: In short, no. But I think I should revisit the pipeline/components which accepts both full path and a folder path of `nii.gz`, and make it so that it accepts the full path to `.nii.gz`

I made the change to `compDcm2nifti` as I think it would make more sense to return the full path from the component itself and not having pipeline to figure out the `.nii.gz` from a folder path.

boonwj: Ok i just understood why it still works. The lung segmentation module assumes the input argument (inputNiftiPath) could either be a directory or a file. 

But even so, this is also not a change within the scope of this PR.

HoloPipelines/pipelines/components/compDicom2numpy.py
boonwj: If GDCM can be removed, then this file should not import pydicom anymore. You will need to clean up the `loadScan()` function and `scan = loadScan(dataPath)`. Don't leave unused code behind.

HoloPipelines/pipelines/components/compSeparateNumpy.py
nielsboecker: If the undesired changes were introduced in a specific commit, you can do a [git revert](https://www.atlassian.com/git/tutorials/undoing-changes/git-revert).

(PS: Without knowing the details, I think I prefer components with the same I/O for all, and making strict constraints rather than accepting a broad range of possible I/O)

HoloPipelines/pipelines/components/compDicom2numpy.py
UdomkarnBoonyaprasert: GDCM is an optional dependency for Pydicom for reading the compressed pixel data. `scan = loadScan(dataPath)` is still needed for reading the individual slice position, and thickness which are used to calculate the resize factor (image re-sampling)

```
        spacing = map(
            float,
            (
                [scan[0].SliceThickness]
                + [scan[0].PixelSpacing[0], scan[0].PixelSpacing[1]]
            ),
        )
        spacing = np.array(list(spacing))
```

HoloPipelines/pipelines/components/compDcm2nifti.py
UdomkarnBoonyaprasert: Yes, I do apologize about that.

nielsboecker: Thanks for resolving this issue. Just for the record: When you continue working on this on another branch, please try to have modular code, as discussed. I.e. ideally, each component should have the same interface, same I/O. I like "dumb" components that have strict constraints what their I/O is over smart components that figure out different input specs depending on the context.

HoloPipelines/pipelines/components/compDicom2numpy.py
boonwj: Thanks for clearing up my misunderstanding. So pydicom can be used without GDCM if you are not dealing with compressed pixel data.

This loc `image = np.stack([s.pixel_array for s in scans])` was the one that required GDCM previously.

ref: https://stackoverflow.com/questions/54657768/pydicom-cant-read-pixel-data-and-needs-gdcm-or-pillow

HoloPipelines/pipelines/components/compDcm2nifti.py
boonwj: I have pushed a commit to undo the changes here. 

The standardisation of I/O of components should be its own PR.

boonwj (commented):
Ok I think the change to Simple ITK looks simple enough

nielsboecker (approved):

PR #28, opened 2019-08-07
Ui/accessor client

nielsboecker:
* use real data from accessor for hologram-related endpoints
* use redirects instead of downloading intermetdiate file and then re-sending it
* `POST` and `DELETE` can be verified with Postman: [>get config<](https://www.getpostman.com/collections/7f4327db4bea1f7a40c3)


fanbomeng97 (approved):
The code is clear and fine, good job!

PR #31, opened 2019-08-08
UI: Integrate with HoloStorageAccessor and HoloPipelines

nielsboecker:
* make some changes to the TS definitions
* implement the POST request to directly upload 3D models to HoloStorageAccessor
* implement the POST request to trigger HoloPipelines generation
  * _(note: the client for this service already assumes that the service is already available, which it is not. It still returns static data at one particular point. Once the Pipelines are up and running, it will ideally only need changing of two lines)_
* redirect all requests to the respective endpoints at the other services


fanbomeng97:
Have reviewed the code, it's very good!
I tried directly upload 3D models, it shows upload successfully, afterwards, there is no uploaded hologram for the  patient, I think it because the related function in HoloStorageAccessor is not ready?

boonwj:
Hmm that should not be the case, as the upload functionality of the accessor is completed already. Could it be that an old accessor version was used?

nielsboecker:
Hmm the ui as it is shown in the browser still has the "fake" success. Did you check the console/network tab if it actually succeeded? The current version of assessor may reject it because dateOfImaging is an empty string. That will soon be changed though

nielsboecker:
I ran the new Accessor 1.1.0 and I can confirm that the functionality works. New holograms get created for upload, and subsequently exist (see screenshot). For generate, it errors because the Pipelines are not ready, but the correct request gets fabricated.
image

~~For some reason the new holograms currently don't show up in the view. Next up I want to hook up the input fields with the values that get send. With this next PR I will also address why the holograms don't show up in the view; this PR can be merged independently.~~
Okay apparently the redirect sets the Origin to null, and Accessor doesn't like it for CORS reasons. Maybe we should actually figure this out before we close the PR.
image

nielsboecker:
**Update**: The main problem was that the CORS headers from the Accessor did not allow redirection as I am performing it now. This will be changed with #36.

I have fixed another small bug and manually verified that I can create new holograms from the UI via Accessor calls now. Currently you have to refresh the page, but then they will show up. Download and deletion work as well.

fanbomeng97 (approved):
Have verified that every thing is working well now, nice work!

PR #32, opened 2019-08-08
Pipeline/abdomen with resize function pipeline

UdomkarnBoonyaprasert:
Adding pipeline to generate glb 3D models. Please note that in order to run the pipeline the segmentation container must be running. The functionality of this pipeline includes: 
- converting Dicom to Nifti
- resize and crop the Nifti (this is actually done during the Numpy array stage using `compNumpyTransformation`) to fit the max resolution for the segmentation model
- generate glb for each organs in the segmentation (8 in total)


boonwj:
What happened with the git history here? There seems to be so many merges and duplicated commits here. 

Also, it's kinda weird that the committer keeps changing between <localadmin> and yourself. I think it might be best to just bring the final state of this branch and commit it as a single commit on top of `PIPELINE/abdomen_pipeline`.

nielsboecker:
Agree, we should try and keep the git history clean.

PR #34, opened 2019-08-09
PIPELINE: Draft version for the API

ansonwong9695:
Created 3 endpoints for the front end
- /job: which receive the post request from the front end, use to activate the pipeline
- /job/<jobid>/status : this endpoint allows query about the status on specific job and return the job status to the front end
- /api/v1/pipelines: this endpoint allows front end to know the information of each pipelines

Create 1 endpoints to update the status internally
- /status: which allow the pipeline update the status dictionary when they finished a section of task

Other functionality
-create the clean up function for the status, if job exist more than 20 mins will be deleted.
-pipeline when finished the task it will send the request to the hologram (have not tested it)


nielsboecker:
Good morning :) Thanks for opening the PR, Kawai. Some general remarks:
* When I meant prefix `/api/v1/`, I meant _for all routes_, not just `/pipelines`. You can read more about this stuff [here](https://docs.microsoft.com/en-us/azure/architecture/microservices/design/api-design), but please change to be all routes.
* Please merge `dev` back into your branch and resolve the merge conflicts.
* Please resolve the CodeFactor issues.

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
     json_file.close()
 
 
+def startPipeline(jobID, plID, paramList=[],infoForAccessor=[]):
+    configFileName = "pipelineList.json"  # hard coded
+    with open(str(pathlib.Path(newCwd).joinpath(str(configFileName)))) as json_file:
```
nielsboecker: semantic naming, maybe just call it `pipelinesConfig`. Actually, why isn't the file called `pipelinesConfig` as well iff it actually contains config?

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
...
+        lsPipe = json.load(json_file)
```
nielsboecker: What does "ls" stand for? why abbreviate?

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
     json_file.close()
 
 
+def startPipeline(jobID, plID, paramList=[],infoForAccessor=[]):
```
nielsboecker: Please make sure to run `black` / the automatic git hooks to ensure a consistent formatting.

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
...
+    configFileName = "pipelineList.json"  # hard coded
+    with open(str(pathlib.Path(newCwd).joinpath(str(configFileName)))) as json_file:
+        lsPipe = json.load(json_file)
+        print(str(["python", lsPipe[plID]["src"], jobID] + paramList+[infoForAccessor]))
```
nielsboecker: Please no "print" in productive code; if it is still needed, use `logging`

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
...
+        #subprocess.run(["python", lsPipe[plID]["src"], jobID] + paramList+infoForAccessor, cwd=newCwd)
```
nielsboecker: Why is this commented out? Isn't this starting the actual processing?

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
...
+        # currently pipelineNifti2glb pipelineDicom2glb are the only pipelines that can take jobID
```
nielsboecker: Hmm which other pipelines need to be changed? Can you please create tasks on the taskboard so we don't lose track?

HoloPipelines/pipelineController.py
```diff
@@ -91,5 +95,25 @@ def main():
...
+
+    json_file.close()
+
+
+def getPipelineList():
```
nielsboecker, line 118: If you have this method, why don't you use it in line 100? Also, again, "ls" and "Pipe" are both unnecessary abbreviations in the variable name

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+import threading
+import time
+import logging
+import requests
+import os
+import pathlib
+from zipfile import ZipFile
+
+app = Flask(__name__)
+app.config["JSON_ADD_STATUS"] = False
```
nielsboecker, line 18: What is `JSON_ADD_STATUS` doing? The name is not expressive. Also to me it seems that this variable is never accessed?

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+# global variables
+this_cwd = pathlib.Path.cwd()
+save_directory = Path("input")
```
nielsboecker: Could be called `input_directory` for consistency

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+# logging formatting
+FORMAT = "%(asctime)-15s -function name:%(funcName)s -%(message)s"
+logging.basicConfig(level=logging.INFO, format=FORMAT)
+
+
+def save_request_file_to_input_dir(filename, request_input_data_URL):
+    global save_directory
+    if not save_directory.is_dir():
+        os.mkdir("input")
```
nielsboecker: Shouldn't hard-code `"input"` here, use the method argument

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+    response = requests.get(request_input_data_URL)
+    if response.status_code != 200:
+        return json_response(jobID="image study endpoint failure", status_code=404)
```
nielsboecker: Status code 404 is okay, but the response should have a different format. There is no jobId.
```
{
    "message": "Failed to fetch imaging study from PACS"
}
```

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+jobid2plid = {"j0": "nifti2glb", "j1": "dicom2glb"}
+job2inputURL = {"j0": "url"}
+job2outputURL = {}
+
+# logging formatting
+FORMAT = "%(asctime)-15s -function name:%(funcName)s -%(message)s"
+logging.basicConfig(level=logging.INFO, format=FORMAT)
+
+
+def save_request_file_to_input_dir(filename, request_input_data_URL):
```
nielsboecker: Could be renamed to `fetch_request_input_file` to make it more expressive

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+def unzip_the_file(filename):
+    # get unzip input filename
+    filename_unzip = filename.rsplit(".", 1)[0]
+    logging.info("unzip file name:" + filename_unzip)
+
+    # unzip the input file
+    file_dir = this_cwd.joinpath(str(save_directory), filename)
+    path_to_store_unzip_file = this_cwd.joinpath(str(save_directory), filename_unzip)
+    with ZipFile(str(file_dir), "r") as zipObj:
+        # unzip
```
nielsboecker: The comments are not necessary here

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+        zipObj.extractall(str(path_to_store_unzip_file))
+    return str(path_to_store_unzip_file)
+
+
+# cleaning the status
+@app.before_first_request
```
nielsboecker: Instead of using this decorator, could you just call this from your `main` method?

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+def activate_status_cleaning_job():
+    def run_job():
+        while True:
+            global status
+            # logging.info("status before loop: " +str(status))
```
nielsboecker: Please remove all the commented log statements. However, it would be good to have one debug statement within the task that runs every 30 sec, that contains some useful information on the current state.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+            for job in status.copy():
```
nielsboecker: why do you `copy()`? This seems like an unneccessary workload and runs every 30 sec...

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+            time.sleep(30)
+
+    # logging.info("status after loop: "+str(status))
+    thread = threading.Thread(target=run_job)
+    thread.start()
+
+
+# update the status from pipeline
+@app.route("/status", methods=["POST"])
+def getTheStatus():
```
nielsboecker: Please make sure to use snake_case consistently, not camelCase

nielsboecker: Why is the method called "get"? If this is used to update, give it an appropriate name like `update_job_status`

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    global status
+    current_job_status = request.get_json()
+    status.update(current_job_status)
+    return json.dumps(status)
```
nielsboecker: This method would benefit from an informational debug statement

nielsboecker: Does it need to always return the whole status for all jobs?? Maybe better just the current one

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+# get pipeline info
+@app.route("/api/v1/pipelines", methods=["GET"])
+def send_list_of_pipelines():
+    global pipeline_list
+    pipeline_dict = {}
+    for (
```
nielsboecker: Why the line breaks and the parenthesis?
Usual syntax is `for key, value in dict.items():`
Also, in your case, instead of `key` and `value`, use `plid` and `pipeline` or something like that.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+        key,
+        value,
+    ) in (
+        pipeline_list.items()
+    ):  # not complete. the value for inputConstraints is wrong (atm it's just number of param)
+        pipeline_dict[key] = {
```
nielsboecker: A couple of things here:
1. You are doing a mapping here. Could extract that to a method like `map_pipelines_info()`.
1. You are doing this in every request to this endpoint, although the pipeline list is static. You could just call it once on the top level.
1. Actually the best thing might be to just _change the format of the pipeline list_ so that you don't have to map it. Just have the 1:1 same keys in the pipeline list json. Then your mapping becomes real easy. It is basically a filtering for the values you want. You send them unchanged, and you jjust use your mapping method to exclude internal fields. See https://stackoverflow.com/a/3420156/8495954

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+            "plid": key,
+            "title": value["name"],
+            "description": value["description"],
+            "inputConstraints": value["inputConstraints"],
+            "inputExampleImageUrl": value["inputExampleImageUrl"],
+            "outputExampleImageUrl": value["outputExampleImageUrl"],
+        }
+
+    return json.dumps(pipeline_dict)  # should we str() here?
```
nielsboecker: No we don't need `str()`, because that's exactly what `json.dumps()` provides (see https://docs.python.org/3/library/json.html) Please remove comment.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+# use to start the pipeline
+@app.route("/job", methods=["POST"])
+def send_job_start_response():
```
nielsboecker: The name should reflect the core thing this method does, for instance `start_job`

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+    if response.status_code != 200:
+        return json_response(jobID="image study endpoint failure", status_code=404)
+    file_dir = this_cwd.joinpath(str(save_directory), filename)
+    open(file_dir, "wb+").write(response.content)
+    logging.info("file dir: " + str(file_dir))
+
+
+def unzip_the_file(filename):
+    # get unzip input filename
+    filename_unzip = filename.rsplit(".", 1)[0]
```
nielsboecker: Do you need the second parameter here?

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+    job_request = request.get_json()
+    request_plid = job_request["plid"]
+    request_input_data_URL = job_request["imagingStudyEndpoint"]
+
+    # get filename
+    if request_input_data_URL.find("/"):
+        filename = request_input_data_URL.rsplit("/", 1)[1]
+        logging.info("filename: " + filename)
+
+    save_request_file_to_input_dir(filename, request_input_data_URL)
```
nielsboecker: This method can just return the `response.content` rather than writing it to a file. Then use that as a paramter to the unzip method, and you have done it in-memory and avoided an additional write to disk: https://stackoverflow.com/questions/2463770/python-in-memory-zip-library

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+    unzip_file_dir = unzip_the_file(filename)
+
+    # create output dir
+    global output_directory
+    if not output_directory.is_dir():
+        os.mkdir("output")
```
nielsboecker: Don't hardocde the string, why do you have the variable?

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # create arglist pass to pipeline controller
+    # (this part will change later, we should not check the pipeline arglist in this way)
+    jobID = str(uuid.uuid1())
+    global jobid2plid
+    jobid2plid.update({jobID: request_plid})
+    if request_plid != "lungDicom2glb":
```
nielsboecker: Duplicated code. Use the same variable for both. Then, do a positive matching, that means for any of the pipelines that _DO_ have the additional param, add it.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+            str(
+                this_cwd.joinpath(
+                    str(output_directory), unzip_file_dir.rsplit("/", 1)[1] + ".glb"
+                )
+            ),
+        ]
+    logging.info("arglist: " + str(arglist))
+
+    #create a list that fetch the response info that needs to post with hologram to the accessor (checking is missing)
+    info_for_accesor={
```
nielsboecker: choose a semantic name that is decoupled from other services, e.g. `meta_data`. To assemble the data, you can do the trick I mentioned earlier: https://stackoverflow.com/a/3420156/8495954

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+    startPipeline(jobID, request_plid, arglist,info_for_accesor)
+    return json_response(jobID=jobID, status_code=202)
+
+
+@app.route("/job/<jobid>/status", methods=["GET"])
+def get_job_status(jobid):
+    if jobid in status:
+        status_for_current_jobid = {jobid: status[jobid]["status"]}
+    else:
+        status_for_current_jobid = {jobid: "does not exist"}
```
nielsboecker: Should not use `jobId`, but `message` as key. Also, do you set the reponse status? should be 200 or 404 respectively.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    return json.dumps(status_for_current_jobid)
+
+
+if __name__ == "__main__":
+    app.run(debug=True, host="127.0.0.1")
```
nielsboecker: Shoud not be debug

HoloPipelines/pipelines/NOTDONEpipelineAbdomenDicom2glb.py
```diff
@@ -0,0 +1,32 @@
+# ----this pl in a nutshell:
```
nielsboecker: Why is this file here? It seems liike it is only a draft, not ready to be merged into `dev`. Also doesn't it belong into #32? Maybe you should revert it for now, as I don't want to have it in the `dev` branch like this, and it doesn't seem related with this PR much.

HoloPipelines/pipelines/components/compConvertDicom2nifti.py
```diff
@@ -0,0 +1,11 @@
+import numpy as np
```
nielsboecker, line 1: Why is this file in the PR? Does it have anything to do with the server?

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
+import requests
+import pathlib
+import json
+from pathlib import Path
+from os import path
+
+def sendFilePostRequest(url, inputFile, outputFile):
```
nielsboecker: Please use snake_case for all methods and parameter names.

nielsboecker: Also please run `black` on all files to format (so there is a space between arguments etc.). I highly recommend using an IDE like PyCharm, it will show you such issues while you code.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    inputFile = str(pathlib.Path(inputFile))
+    outputFile = str(pathlib.Path(outputFile))
+    file = {"file": open(inputFile, "rb")}
```
nielsboecker: use `with open...` for the file handling

nielsboecker: Also, use more specific names like `input_file`. Don't use the same variable for something else in line 12.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    response = requests.post(url, files=file)
+    file = open(outputFile, "wb")
+    file.write(response.content)
+    file.close()
+    returnCode = response.status_code
+    print(returnCode)
```
nielsboecker: Use logging. Also you may not need a `returnCode` variable, just use the `response.status_code` value inline

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+):  # TODO rename this comp to just http request? and rename this function to updateJobStatus
+    pipelineServerURL = "http://localhost:5000/status"
+    data = {jobID: {"status": jobStatus, "timestamp": str(datetime.now())}}
+    # data = {"jobID": "j0", "status": "hey", "stamp": "2020"}
+    response = requests.post(url=pipelineServerURL, json=data)
+    returnCode = response.status_code
+    print(returnCode)
+
+
+def sendFilePostRequest(url, inputFile, outputFile):
```
nielsboecker: Duplicate code from `compHttpRequest`?

nielsboecker: Also, the method is never called? The other one in `compHttpRequest` is also never called?

HoloPipelines/pipelines/components/getMultipleNumpy.py
```diff
@@ -0,0 +1,50 @@
+#TODO change file name cos i legit dont know what to name this. File naming is not the same to the rest of comp format anyway
```
nielsboecker: Renamed to `HoloPipelines/pipelines/components/compSeparateNumpy.py ` in Pap's branch - you guys should probably merge your stuff or you will get into trouble if you each make changes in now different files

nielsboecker (changes requested):
As discussed

HoloPipelines/pipelines/components/test_accessor.py
```diff
@@ -0,0 +1,25 @@
+from compHttpRequest import sendFilePostRequestToAccessor
```
nielsboecker: As this is a test, it should probably be in the `test` folder

HoloPipelines/pipelines/components/test_accessor.py
```diff
@@ -0,0 +1,25 @@
...
+
+url="http://localhost:3200"
+directory="/home/kawai/git/HoloRepository-Core/HoloPipelines/output/left-scfe-pelvis-bone.glb"
```
nielsboecker: Hard-coded, this won't work on any other machine. Same for the localhost -- when this test runs on the build server, it will fail because there is nothing at `localhost:3200`. Also you will repeatedly create actual new data in the database if you don't mock, but do a real call to the Accessor.

Not sure what to do with this. Any ideas @boonwj? Maybe [skip the test in automatic runs](https://stackoverflow.com/questions/38442897/how-do-i-disable-a-test-using-py-test), rename it to `accessor_integration_test`, and add documentation that this can only be run manually.

HoloPipelines/pipelines/components/test_accessor.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": {
+              "full":"Andy Lau",
+              "title": "Mr",
+              "given": "Andy",
+              "family": "Lau"
+            }
+  }
+
+
+sendFilePostRequestToAccessor("pelvis generate from dicom to glb",url,directory,"pelvis bone gernerate from dicom to glb pipeline","hip","2017-07-15T15:20:25Z","2017-07-15T15:20:25Z","pelvis generate from dicom2glb",author,patient)
```
nielsboecker: This test is not really testing anything. No asserts, just a shot-in-the-dark method call. In case we keep this test around as a manual integration test or whatever: You could return the response code from the mmethod, and in your test verify that it was successful.

(or better, return the _body_ of the response. You know, a successful POST to the Accessor returns the JSON of the created hologram. then you can actually verify that it created the right data, same aid, name, etc)

HoloPipelines/pipelines/components/compHttpRequest.py
nielsboecker: Is this dead code? Seems you never call it.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    return outputFile
+
+
+def sendFilePostRequestToAccessor(title,url, outputFileDir,description,bodySite,dateOfImaging,creationDate,creationDescription,author,patient):
+    outputFileDir = str(pathlib.Path(outputFileDir))
+    sizeOfOutputFile=str(path.getsize(outputFileDir)/1000)
```
nielsboecker: Should be 1024?

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    file = open(outputFile, "wb")
+    file.write(response.content)
+    file.close()
+    returnCode = response.status_code
+    print(returnCode)
+    return outputFile
+
+
+def sendFilePostRequestToAccessor(title,url, outputFileDir,description,bodySite,dateOfImaging,creationDate,creationDescription,author,patient):
+    outputFileDir = str(pathlib.Path(outputFileDir))
```
nielsboecker: Again, please use snake_case for variable names. You should install an IDE or some plugin for your text editor so that it does a check against PEP8 style guide and warns you before you commit.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    sizeOfOutputFile=str(path.getsize(outputFileDir)/1000)
+    file = {"file": open(outputFileDir, "rb")}
+    #print("file: "+str(file)
```
nielsboecker: Pls remove

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    # manipulate patient data
+    patientForAccessor={
+        "pid": patient["pid"],
+        "name": patient["name"]
+    }
+
+
+    requestBody = {
```
nielsboecker: This will need to be changed to form data and multifor , but I think you are already working on that.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+        "creationDescription":creationDescription,
+        "hologramFile":[file],
+        "author": authorForAccessor,
+        "patient": patientForAccessor
+
+        }
+    response = requests.post(url, data=requestBody)
+
+    returnCode = response.status_code
+    print(returnCode)
```
nielsboecker: Use logging instead, and maybe provide some context instead of just the number. Also, the response's body is interesting as well.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    return requestBody
```
nielsboecker: Why would you return the requestBody? I think you either want to return a boolean (whether the response was successful) ~~or maybe the response body (but that might be unnecessary if you don't acually use it anywhere)~~.

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
+import requests
```
nielsboecker: I don't like that every component that calls this component has their own, free-style strings to set their status. I would prefer if you had either an enum or a couple of constant string variables here in this file, that other components reference for the `jobStatus` argument)

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+from datetime import datetime
+import pathlib
+
+
+def updateStatus(
+    jobID, jobStatus
+):  # TODO rename this comp to just http request? and rename this function to updateJobStatus
```
nielsboecker: No, it is okay to have status in one component. You could have another component that just does the http requests. And in the status component, you just call the other one (so you don't even know the pipeline server url here in this one)

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    pipelineServerURL = "http://localhost:5000/status"
```
nielsboecker: Please change 5000 to 3100 everywhere, as discussed. Ideally, it should be configurable and read from a separate config file.

nielsboecker: Please add documentation why we are doing this. And maybe rename to `self_url` to make it more clear what is going on.

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    data = {jobID: {"status": jobStatus, "timestamp": str(datetime.now())}}
+    # data = {"jobID": "j0", "status": "hey", "stamp": "2020"}
```
nielsboecker: Pls remove

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    response = requests.post(url=pipelineServerURL, json=data)
+    returnCode = response.status_code
+    print(returnCode)
```
nielsboecker: Again, please use logging, and write some context to it. Also no need to store it in a separate variable.

HoloPipelines/pipelines/components/compJobStatus.py
nielsboecker: But if this component here is responsible for keeping the status, then you should also move the dict of the status inside this file. And then from other files, you can import it to access it.

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    with open(inputFile, "rb") as f:
+        response = requests.post(url, files={inputFile: f})
+    file = open(outputFile, "w+")
+    file.write(response.content)
+    file.close()
+    returnCode = response.status_code
+    print(returnCode)
+
+
+def sendGetRequest(url, requestBody):
```
nielsboecker: 1. I think is never used as well??
1. Please put only status related methods inside this file and remove unrelated stuff.

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    response = requests.post(url=url, json=requestBody)
+    returnCode = response.status_code
+    print(returnCode)
```
nielsboecker: Print statement

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+if __name__ == "__main__":
+    updateStatus("j0", "hello")  # TODO remove once done testing
```
nielsboecker: 😐

HoloPipelines/pipelines/components/compJobStatus.py
```diff
@@ -0,0 +1,39 @@
...
+    print(
+        "You shouldn't be able to run this component directly after we're done testing"
+    )  # TODO look at this too pls
```
nielsboecker: What does this mean? Such comments should be removed before merging this to `dev`

HoloPipelines/pipelines/components/compJobStatus.py
nielsboecker: This should not be here. Instead you can write an actual test for this component (after moving the actual state here), setting state , and then verifying that it is set. Easy as that.

HoloPipelines/pipelines/components/compHttpRequest.py
```diff
@@ -0,0 +1,73 @@
...
+    response = requests.post(url, files=file)
+    file = open(outputFile, "wb")
+    file.write(response.content)
+    file.close()
+    returnCode = response.status_code
+    print(returnCode)
+    return outputFile
+
+
+def sendFilePostRequestToAccessor(title,url, outputFileDir,description,bodySite,dateOfImaging,creationDate,creationDescription,author,patient):
```
nielsboecker: These are too many arguments. Consider redisgning it or [introducing a parameter object](https://softwareengineering.stackexchange.com/a/145066)

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -1,14 +1,18 @@
...
 from components import compNumpy2obj
 from components import compObj2glbWrapper
+from components import compHttpRequest
+from datetime import datetime
 import pathlib
 import sys
 
 
-def main(dicomPath, outputGlbPath):
+def main(jobID, dicomPath, outputGlbPath,infoForAccessor):
```
nielsboecker: `metaData` instead of `infoForAccessor`

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -19,19 +23,35 @@ def main(dicomPath, outputGlbPath):
...
     )
+
+    compJobStatus.updateStatus(jobID, "3D format conversion")
     generatedGlbPath = compObj2glbWrapper.main(
         generatedObjPath,
         str(pathlib.Path(outputGlbPath)),
         deleteOriginalObj=True,
         compressGlb=False,
     )
     print("lungDicom2glb: done, glb saved to {}".format(generatedGlbPath))
```
nielsboecker: Use logging instead of print

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -19,19 +23,35 @@ def main(dicomPath, outputGlbPath):
...
+    compJobStatus.updateStatus(jobID, "Finished")
+    compHttpRequest.sendFilePostRequestToAccessor(
+        infoForAccessor["bodySite"]+"apply on lung segmentation",
```
nielsboecker: Missing a space here. Also the body site is pretty obvious I guess :) How about just "Lung segmentation"?

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -19,19 +23,35 @@ def main(dicomPath, outputGlbPath):
...
+        "http://localhost:3200/api/v1/holograms",
```
nielsboecker: The URL should not be hard-coded; it could be another domain, once we deploy to production. So please use a config file. Also, the URL should not be aparameter for this method, because it's always going to the accessor. It should be kept inside the http component

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -19,19 +23,35 @@ def main(dicomPath, outputGlbPath):
...
+        infoForAccessor["description"],
+        outputGlbPath,infoForAccessor["bodySite"],
+        infoForAccessor["dateOfImaging"],
+        datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
```
nielsboecker: creationDate should not be a paramter for this method, as it will be dynamically created all the same. Should be done in http component.

HoloPipelines/pipelines/pipelineLungDicom2glb.py
```diff
@@ -19,19 +23,35 @@ def main(dicomPath, outputGlbPath):
...
+        "Generate glb mesh from dicom",
```
nielsboecker: For consistency, how about having the parameter just be the pipeline name? Then in the http component, the creationDescription could always be `f"Generated with {pipelineName] pipeline."`

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,202 @@
...
+
+            time.sleep(30)
+
+    # logging.info("status after loop: "+str(status))
+    thread = threading.Thread(target=run_job)
+    thread.start()
+
+
+# update the status from pipeline
+@app.route("/api/v1/status", methods=["POST"])
```
nielsboecker: There are actually more elegang ways to have the same prefix for all routes, which also makes it easier if it changes. See https://stackoverflow.com/questions/18967441/add-a-prefix-to-all-flask-routes

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,202 @@
+from flask import Flask, request
```
nielsboecker, line 1: This file does way too much stuff which is not server related. Please do at least extract other components:
* to keep the status of jobs
* to do the downloading and unzipping stuff

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,202 @@
...
+app = Flask(__name__)
+app.config["JSON_ADD_STATUS"] = False
+
+# global variables
+this_cwd = pathlib.Path.cwd()
+save_directory = Path("input")
+output_directory = Path("output")
+pipeline_list = getPipelineList()
+status = {
+    "j0": {"status": "segment", "stamp": "2019-08-05 14:09:19"},
```
nielsboecker: Ummm??? This should probably be removed?? You want to start with an empty status dictionary.

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,202 @@
...
+    "j1": {"status": "segment", "stamp": datetime.now().strftime("%Y-%m-%d %H:%M:%S")},
+}
+jobid2plid = {"j0": "nifti2glb", "j1": "dicom2glb"}
```
nielsboecker: None of these three dictionaries are needed. job to plid, job to inputURL, job to outputURL. Please remove all.

nielsboecker (changes requested):
Please follow up with the comments, feel free to ask if you have any questions. Before we can accept the PR, we should also verify that
- the status updates are actually working (can be verified if you add debug statements)
- the data that gets created for a job will also be cleaned after it's finished
- the POST to the Accessor is actually working

nielsboecker:
@ansonwong9695, @UdomkarnBoonyaprasert: I have some other TODOs for you guys; we can track them in this post and cross them out when you finished them:

- [x] I see an architectural problem: Initially, we said both the data fetching from PACS and the data posting to Accessor should be their own components, that can be part of each pipeline. In your code however, the fetching gets done right in the server, whereas the posting gets done at another level (in each pipeline). Please refactor.
- [ ] ~~You need to clean up after yourselves. What happens with all the input files you download after the job is finished? They will stay there forever, currently. Also they will all be in the same directory, it is impossible to keep track what came from where. Potentially there can even be issues if different pipelines try to download a file with the same name at the same time. How about creating a folder `jobs/<jobid>` for each job. `jobs/<jobid>/input` for your stuff from the PACS and `jobs/<jobid>/output` for what you are generating; and then maybe deleting that once stuff gets removed from the status? @boonwj any other ideas?~~ NOT done. Shall be done in a new PR.
- [x] You should add flask to your `dependencies.txt`
- [ ] ~~You should update the `README` on how to start the server and document the endpoints.~~ I do it in the refactoring branch
- [x] There are two empty files, please remove: pipelines/components/compStoreGLB and pipelines/components/compOutputGLB
- [ ] ~~You should add a `Dockerfile` (this can be a separate PR, but don't forget it) => different PR~~
- [ ] ~~There should be more error handling (you always expect that all fields are coming exactly as expected, but what happens otherwise?)~~ - can be a separate PR

HoloPipelines/pipelineServer.py
nielsboecker, line 1: Rather than having them on the top level, could you move this file and `pipelineController` to a new directory/module `HoloPipelines/core`

HoloPipelines/pipelineServer.py
ansonwong9695: because I use the flask_json library for creating json response to the front end and in default it will generate a status code by default but the status code i want to generate is not the default value. for example in default  library will return a json response to the request with status code 200 but i want to send 202 instead.

nielsboecker: Ohh okay. Please add flask_json to the requirements.txt then, along with other dependencies. 

Btw, if i understand the docs correctly, it will just include the status code _that you set_, not always the default value. If you do `return json_response(jobID=jobID, status_code=202)`, it should include 200. But it's fine to just turn the feature off, I guess.

HoloPipelines/pipelineServer.py
ansonwong9695: because i am modifying the status dict inside a loop so i need to create a copy of the status

HoloPipelines/pipelineServer.py
ansonwong9695: yes, it only returns the current one

nielsboecker: Oh so you mean you changed it? Because the code as it currently is commited here looks like you return `status` (all) instead of `current_job_status`

HoloPipelines/pipelineServer.py
nielsboecker: Sorry I don't understand. Why do you have to copy it then? What's the problem with the loop? What will break if you do ` for job in status: ...`

ansonwong9695: RuntimeError: dictionary changed size during iteration
this is the error i will get if i do not use copy, because in this for loop i will delete the key and value inside this dictionary so the size is going to change

HoloPipelines/pipelines/components/getMultipleNumpy.py
nielsboecker: deleted

HoloPipelines/pipelineServer.py
nielsboecker: Hm not sure, will leave it for now as the dict will be replaced most likel

HoloPipelines/pipelines/NOTDONEpipelineAbdomenDicom2glb.py
nielsboecker: deleted

HoloPipelines/pipelines/components/compStatus.py
```diff
@@ -0,0 +1,9 @@
+from datetime import datetime
+
+status = {
```
nielsboecker, line 3: Why are there still things in here? Shouldn't this start out empty???

HoloPipelines/pipelines/components/compJobStatus.py
nielsboecker: Was marked as resolved but comment still here. I will remove it now.

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+    if not os.path.exists(str(job_path)) and create_dir:
+        os.mkdir(str(job_path))
+    return job_path
+
+
+def str_job_path(job_ID):
+    return str(pathlib_job_path)
+
+
+def make_str_job_path(job_ID, sub_dir_list, create_sub_directories=True):
```
nielsboecker, line 20: Method name implies that it just returns a string, but it creates directories

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
+import pathlib
+import shutil
+import os
+import logging
+
+this_comp_path = str(pathlib.Path(str(os.path.dirname(os.path.realpath(__file__)))))
+
+
+def pathlib_job_path(job_ID, create_dir=True):
```
nielsboecker, line 9: Why "pathlib" in method name?

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+    job_path = pathlib.Path(this_comp_path).parents[1].joinpath("jobs", str(job_ID))
+    if not os.path.exists(str(job_path)) and create_dir:
+        os.mkdir(str(job_path))
+    return job_path
+
+
+def str_job_path(job_ID):
```
nielsboecker, line 16: Unused parameter `job_ID`

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+def make_str_job_path(job_ID, sub_dir_list, create_sub_directories=True):
+    if (
+        not os.path.isdir(str(pathlib_job_path.joinpath(*sub_dir_list).parent))
+        and create_sub_directories
+    ):
+        os.makedirs(str(pathlib_job_path.joinpath(*sub_dir_list).parent))
+    return str(pathlib_job_path.joinpath(*sub_dir_list))
+
+
+def clean_up(job_ID):
```
nielsboecker, line 29: Unused parameter `job_ID` -- This can never work. Have you seen this work for any actual job???

HoloPipelines/pipelines/components/compJobPath.py
nielsboecker, line 20: Unused parameter `job_ID` --- How can this even work???

nielsboecker: Again, let me advise you strongly to use an IDE, as it helps:
image

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+    if os.path.exists(str_job_path):
```
nielsboecker, line 30: `str_job_path` is the method name, but you are not calling it?

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+
+def pathlib_job_path(job_ID, create_dir=True):
+    job_path = pathlib.Path(this_comp_path).parents[1].joinpath("jobs", str(job_ID))
+    if not os.path.exists(str(job_path)) and create_dir:
+        os.mkdir(str(job_path))
+    return job_path
+
+
+def str_job_path(job_ID):
+    return str(pathlib_job_path)
```
nielsboecker, line 17: `pathlib_job_path` is the method name, but you are not calling it?

nielsboecker: Why would you even need a method that just does a str() of something? Also why don't you just return a str() of thhe Path in line 13?

HoloPipelines/pipelines/components/compJobPath.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def make_str_job_path(job_ID, sub_dir_list, create_sub_directories=True):
+    if (
+        not os.path.isdir(str(pathlib_job_path.joinpath(*sub_dir_list).parent))
```
nielsboecker, line 22: In line 22-26 you again use `pathlib_job_path` like a variable, but it is a method,, and you are not calling it? Also I don't understand what is happening iin these lines. Please clarify and/or write comments.

HoloPipelines/pipelines/components/compGetInput.py
```diff
@@ -0,0 +1,31 @@
+import io
+import requests
+from pathlib import Path
+from flask_json import json_response
+from zipfile import ZipFile
+import logging
+
+input_directory = Path("input")
```
nielsboecker, line 8: How is this supposed to be the job-specific dir??

HoloPipelines/pipelineConfig.py
```diff
@@ -0,0 +1,140 @@
...
+    # process.join
+    # dynamicLoadingPipeline(pipeline_ID, dictionary)
+
+
+def worker():
+    while True:
+        logging.debug("This is logging for TEST1")
+
+
+def dynamicLoadingPipeline(plID, parameter_dict):
```
nielsboecker, line 122: This method is duplicate here  and in pipelineController and both seem to be unused? @ansonwong9695

HoloPipelines/pipelineController.py
```diff
@@ -40,56 +48,93 @@
...
         subprocess.run(
-            ["python", lsPipe[args.pipelineID]["src"]] + args.param, cwd=newCwd
+            ["python", list_of_pipeline[args.pipelineID]["src"]] + args.param,
+            cwd=new_cwd,
         )
 
     json_file.close()
 
 
+def startPipeline(pipeline_ID, parameter_dict):
```
nielsboecker, line 106: Duplicate code and seems to never be called??

HoloPipelines/pipelineController.py
```diff
@@ -40,56 +48,93 @@
...
+def getPipelineList():
+    configFileName = "pipelineList.json"  # hard coded
+
+    with open(str(pathlib.Path(new_cwd).joinpath(str(configFileName)))) as json_file:
+        list_of_pipeline = json.load(json_file)
+    json_file.close()
+    return list_of_pipeline
+
+
+def dynamicLoadingPipeline(plID, parameter_dict):
```
nielsboecker, line 127: Duplicate code and seems to never be called?

HoloPipelines/pipelineServer.py
```diff
@@ -0,0 +1,108 @@
...
+import pathlib
+
+app = Flask(__name__)
+app.config["JSON_ADD_STATUS"] = False
+PREFIX = "/api/v1"
+
+# global variables
+this_cwd = pathlib.Path.cwd()
+
+output_directory = Path("output")
```
nielsboecker, line 24: How is this supposed to be the job specific directory?

nbckr:
[CodeFactor](https://www.codefactor.io/repository/github/nbckr/holorepository-core) found an issue: Duplicate Code

It's currently on:
[HoloPipelines\pipelineConfig.py:29-82
](https://github.com/nbckr/HoloRepository-Core/blob/d23976d68e06445195e559343b1c6743b133cec3/HoloPipelines/pipelineConfig.py#L29-L82)[HoloPipelines\pipelineController.py:26-79
](https://github.com/nbckr/HoloRepository-Core/blob/d23976d68e06445195e559343b1c6743b133cec3/HoloPipelines/pipelineController.py#L26-L79)Commit d23976d68e06445195e559343b1c6743b133cec3

HoloPipelines/pipelineConfig.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+parser = argparse.ArgumentParser(description="Select pipeline to process")
+parser.add_argument(
+    "-c",
+    "--config",
+    default="pipelineList.json",
+    type=str,
+    help="path to pipeline config file relative to pipelineConfig",
+)
```
nielsboecker, line 29: Duplicate Code

It's currently on:
HoloPipelines\pipelineConfig.py:29-82
HoloPipelines\pipelineController.py:26-79

nielsboecker (changes requested):
@ansonwong9695 Thanks for cleaning up.
* Most things are okay now; i cleaned up some small stuff myself
* Please go through the few comments relevant for your code sections
* Please confirm that everything stll works (or fix if i introduced small bugs)
* ~~We will push the Dockerfile and the logging to files feature to later (not in this PR), i created stories on the task board~~
* ~~there are some issues in CodeFactor left; I think we can leave them there as these components wil hopefully be refactored~~

@UdomkarnBoonyaprasert You told me the clean-up functionality was finished, so now I am thoroughly confused. The clean up component can never have worked, because of crucial errors in there?
* please fix these and all other comments
* please verify and confirm here that the clean up actually worked for you
* please verify that my merge hasn't broken your code
* ~~ideally the clean up should have tests but there is no time, so screw that~~

@~~UdomkarnBoonyaprasert and @ansonwong9695: There are also conflicts for many files from `dev`, so that's another thing we will have to fix...~~ Done by me.

nielsboecker:
Sorry @ansonwong9695, I realised that the build for your work was broken all along. I fixed several issues with the imports, but the tests still fail. I think  it is because you changed the `main` method signature (using the object to bundle all the arguments). But you did not change the tests to reflect that. Can you please update the tests?

nielsboecker:
**Update**: I found that the tests fail for different reasons actually, including @UdomkarnBoonyaprasert's changes. It's a big, big, mess to be honest. I disabled the tests for now, and you can fix them later (in a different branch!)

HoloPipelines/pipelines/components/compGetInput.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelineConfig.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelineController.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelineController.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelineServer.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelineConfig.py
nielsboecker: I'll fix it in refactoring branch.

HoloPipelines/pipelines/components/compStatus.py
nielsboecker: I'll fix it in refactoring branch.

nielsboecker (approved):

nielsboecker:
Note that there are still many issues in this branch at the time of merging, especially in `compJobPath`. I am merging it now to prevent future merge conflicts, as I am doing a heavy refactoring in #45 . In that branch I also fix most issues. The "compJobPath" issues can be solved in #59.

PR #36, opened 2019-08-09
STORAGE: Access-Control-Allow-Origin: *

nielsboecker:
This is needed to allow redirection from the UI.

The current solution to redirect all hologram-related requests directly to the Accessor is not ideal and might be refactored if time allows. However, at least for the `/download` endpoint, this solution makes sense. So for now, I suggest we do `Access-Control-Allow-Origin: *` for the Accessor.


boonwj (approved):
It's a simple change. I have extended it to use a configuration field to enable the functionality.

PR #38, opened 2019-08-10
MISC: Add docker-compose

nielsboecker:
Start the whole system with docker-compose 🚀


fanbomeng97 (approved):
Have verified that everything works well.

PR #41, opened 2019-08-13
STORAGE/Refactor & Testing

boonwj:
This PR brings the following changes to the HoloStorageAccessor

## Refactor
- Shift configuration and blob initialisation out of the router setup function.
- Configuration will be passed in to the setup function

## Tests
- Create mock FHIR server for testing of responses 
- Add tests for all GET and PUT endpoints in HoloStorageAccessor 

Code coverage currently at ~70%. Lacking tests for blobstorage related code (e.g. Hologram DELETE / POST, blobstorage operations)


HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
```diff
@@ -0,0 +1,1470 @@
...
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// This creates a test server that mocks the FHIR server responses
```
nielsboecker: Maybe a separate file?

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
```diff
@@ -0,0 +1,1470 @@
...
+							"given": [
+								"Britney"
+							],
+							"prefix": [
+								"Ms"
+							]
+						}
+					],
+					"gender": "female",
+					"birthDate": "1960-01-02"
```
nielsboecker: Damn Britney is old

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
```diff
@@ -0,0 +1,1470 @@
...
+	}
+
+	tests := map[string]test{
+		"no_aids_query": {
+			inUrl:      "/api/v1/patients",
+			inMethod:   http.MethodGet,
+			wantStatus: 400,
+		},
+		"wrong_query_field": {
+			inUrl:      "/api/v1/patients?pid=p1000",
```
nielsboecker, line 221: Nice

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
```diff
@@ -0,0 +1,1470 @@
...
+			inMethod:   http.MethodGet,
+			wantStatus: 404,
+			wantErr: Error{
+				ErrorCode:    "404",
+				ErrorMessage: `id 'a3000' cannot be found`,
+			},
+		},
+	}
+
+	// Start a local HTTP server and Router
```
nielsboecker: I don't understand why you do this eight times?

nielsboecker (approved):
Looks great and the test coverage now really provides some serious confidence.

The only thing I'm not sure about is that `api_default_test.go` is really long. Maybe it could be broken apart by main routes, e.g. `api_default_holograms_test.go`. The mock FHIR server could be its own file and for each test, only the relevant routes need to be mocked with return values.

We could then also consider to put all of that stuff in a `test` folder, even if it makes imports a bit harder and breaks the golang tradition of keeping it flat? To make it more concise you could even think about extracting the expected JSONs into actual `json` files.

But having said that I'm not sure if it is a good idea as I don't know the language, and I don't think you should put much more time into it necessarily.

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
boonwj: shifted it to a separate file `mock_server_test.go`

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
boonwj: absolutely

HoloStorageAccessor/internal/holostorageaccessor/api_default_test.go
boonwj: Changed it such that `TestMain()` function does an initial setup before running all tests.

boonwj:
Addressed most of the concerns except for shifting the code out of the flat structure and moving the test input to jsons.

 Also found out that I have some bloated code that was removed.

nielsboecker (approved):
Double-approve

PR #44, opened 2019-08-13
UI: Use the real form data to build POST requests and improve form UX

nielsboecker:
* Add a library to do form validation and only allow the user to continue when all required fields are set
  * _(Note: The solution is not ideal. It is a bit of an overkill and there are probably easier, more lightweight solutions. But I think it's good enogh, it is still clean design, and it works)_
* Adjust tests (the library is a Higher-Order-Component and therefore needs wrapping in tests)
* Update the view after a hologram was uploaded or deleted, so the view now actually reflects what's going on (no manual refresh needed)
* Redirect after uploading was successful (earlier, this was mocked)
* Various small improvements and refactorings


fanbomeng97 (approved):
Have verrified that everything works well, the code is also good, nice work.

PR #52, opened 2019-08-15
UI: Extract config

nielsboecker:
This PR will extract config from UI client and server to env files and therefore make it easy to configure in deployment.

**Note**: Due to my PC problems, I could not test the Docker-related changes.


fanbomeng97 (approved):
Have verified that everything works well, nice work!

Issue #47, opened 2019-08-14
PIPELINE: Perform downscaling for all pipelines


nielsboecker:
# Description
Currently, we are only scaling down in the abdominal organs pipeline. For that one, we need to do it, because the NN will crash with larger inputs. The assumption so far was that for other pipelines we can accept all input sizes.

However, this causes performance issues. For example, the bone segmentation takes a lot longer with large inputs. In a test run, the marching cube step of that pipeline (`compNumpy2Obj`) took 45 sec.

Therefore, the downscaling component should be included in all pipelines and make sure the input will be scaled down to a maximum size. For the concrete values, we should try different values, and find a sweet spot between performance and quality.

nielsboecker:
Changes made by @UdomkarnBoonyaprasert in #54, which I then transfered to #45

Issue #45, opened 2019-08-14
PIPELINE: Major refactoring needed [WIP]


nielsboecker:
# Description
Our HoloPipelines code base has some major flaws, which already lead to architectural and maintanance-wise problems:

* we barely distinguish between what we are doing ("resample an image") and technical details, which are not necessarily related to the tas ("dicom2numpy")
* similarly, the pipelines that are physically in our code ("nifty2glb") do not obviously reflect the semantic use cases that we offer to the user ("Bone segmentation pipeline")
* we are misusing the concept of "components", as everything is now just a component
* we are inconsistent in our naming conventions and often break Python style guides

## Suggested solution
Major refactoring! :rocket:
* A new naming scheme: `foo_bar.task.py` instead of `compFooBar.py`
* Different types of components:
  * `tasks` instead of components
  * `services` for more complex stuff, like format conversion
  * `wrappers` for external tools and existing solutions
  * `clients` for communications like HTTP
  * `adapters` to abstract similar http calls, e.g. to the NiftyNet API we defined in the `/models` directory
* Clear separation of concerns for what used to be "components"
  * `pipelines` are made up entirely from `tasks`
  * `tasks` can use `wrappers` for external services, `services` for complex code, `clients` if needed
  * `tasks` can implement some functionality directly inside them if appropriate, but it is preferred to extract as much as possible to other types of components 
* Rename methods and variables to have `snake_case`

## Suggested new structure

```python
/core/
   pipelines.controller.py                  # Only for direct local invocation
   main.server.py                           # Just to listen to endpoints and then
                                            # delegate to other files, like
                                            # jobs.controller or jobs_status
   jobs.controller.py                       # Start pipelines in production
   jobs_status.py                           # Keep track of jobs (could maybe
                                            # be kept in jobs.controller instead)
   garbage_collector.py                     # Perform the "clean up" for jobs

   
   /pipelines/
      bone_segmentation.pipeline.py
      abdominal_organs_segmentation.pipeline.py
      lung_segmentation.pipeline.py
      airway_segmentation.pipeline.py
      README.md                             # Descriptions of each pipeline, like 
                                            # in HoloPipelines/models/README.md
     
      /config/
         pipelines_list.json
         io_paths.py
         config.ini
     
      /tasks/
         /shared/
            receive_input.task.py
            dispatch_output.task.py
            resample_numpy.task.py
            rotate_numpy.task.py
            crop_numpy.task.py
            downscale_numpy.task.py
[...]

PR #54, opened 2019-08-16
PIPELINE: rescaling for rest of the pipelines

UdomkarnBoonyaprasert:
- add resize function call from compNumpyTransforamtion in the rest of the pipelines (abdomen already has this)
- all pipelines except for abdomen pipeline now has resolution limit of 256
- add logging info for marching cube (time taken and face count)
- minor fix with dicom rotation

This is to address issue from #47.

### DIFF
--- a/HoloPipelines/pipelines/components/compDicom2numpy.py
+++ b/HoloPipelines/pipelines/components/compDicom2numpy.py
@@ -88,7 +88,7 @@ def reorientateNumpyList(numpyList):
     # transpose numpy i.e. (z, y, x) ---> (x, y, z)
     numpyList = numpyList.transpose(2, 1, 0)
     numpyList = np.flip(numpyList, 0)
-    numpyList = np.flip(numpyList, 1)
+    # numpyList = np.flip(numpyList, 1) # need to confirm this
     return numpyList
 
 
--- a/HoloPipelines/pipelines/components/compDicom2numpy.py
+++ b/HoloPipelines/pipelines/components/compDicom2numpy.py
@@ -88,7 +88,6 @@ def reorientateNumpyList(numpyList):
     # transpose numpy i.e. (z, y, x) ---> (x, y, z)
     numpyList = numpyList.transpose(2, 1, 0)
     numpyList = np.flip(numpyList, 0)
-    # numpyList = np.flip(numpyList, 1) # need to confirm this
     return numpyList
 
 
--- a/HoloPipelines/pipelines/pipelineDicom2glb.py
+++ b/HoloPipelines/pipelines/pipelineDicom2glb.py
@@ -2,17 +2,21 @@ from components import compCommonPath
 from components import compDicom2numpy
 from components import compNumpy2obj
 from components import compObj2glbWrapper
+from components import compNumpyTransformation
 import pathlib
 import sys
 import logging
+import datetime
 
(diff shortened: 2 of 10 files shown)


HoloPipelines/pipelines/components/compDicom2numpy.py
```diff
@@ -88,7 +88,7 @@ def reorientateNumpyList(numpyList):
     # transpose numpy i.e. (z, y, x) ---> (x, y, z)
     numpyList = numpyList.transpose(2, 1, 0)
     numpyList = np.flip(numpyList, 0)
-    numpyList = np.flip(numpyList, 1)
+    # numpyList = np.flip(numpyList, 1) # need to confirm this
```
nielsboecker: Can you confirm this today as you will be unavailable for the next 6 days? Otherwise it's hard to accept this PR

HoloPipelines/pipelines/pipelineDicom2glb.py
```diff
@@ -2,17 +2,21 @@
...
 import sys
 import logging
+import datetime
 
 logging.basicConfig(level=logging.INFO)
 
 
 def main(dicomFolderPath, outputGlbPath, threshold):
     generatedNumpyList = compDicom2numpy.main(str(pathlib.Path(dicomFolderPath)))
+    resizedNumpyList = compNumpyTransformation.sizeLimit(generatedNumpyList, limit=256)
```
nielsboecker: if 256 is limit for all pipelines, or at least the default, maybe we can bring it to the shared configuration file?

nielsboecker (changes requested):

HoloPipelines/pipelines/pipelineDicom2glb.py
nielsboecker: How did you decide for 256 for all others, and why does the abdomen pipeline have a different value?

HoloPipelines/pipelines/components/compDicom2numpy.py
UdomkarnBoonyaprasert: Confirmed, NumPy array is now in correct orientation. Commented code removed

HoloPipelines/pipelines/pipelineDicom2glb.py
UdomkarnBoonyaprasert: I added a `config.ini` in the base level of HoloPipeline. At the moment it only holds default value for resolution limit. This value is called as default value in `compNumpyTransformation`

Resolution limit of abdomen pipeline is now set to default (256). I chose 256 for default as the final outputs from experiment did not show significant downgrade in quality, also that this value is approximately the median to the values I usually get from most Dicom after resampling (220 to 300). We also plan to implement fast quadratic simplification (wrapper component still in my local branch) we can expect the file to be smaller.

HoloPipelines/pipelines/components/compDicom2numpy.py
nielsboecker: thanks

nielsboecker:
**Update**: Thanks @UdomkarnBoonyaprasert. As discussed, I read your changes and copied them to my branch in  80069112e7b59319214b9db305074976a3601e41 for #45. I will therefore reject this PR.

PR #63, opened 2019-08-21
Ui/remove deprecated docker compose

boonwj:
This PR removes the deprecated docker-compose file in the UI. It also updates the main docker-compose file to use the local UI configurations.


README.md
```diff
@@ -165,6 +165,13 @@ Starting holorepository-core_holorepository-ui-server_1                      ...
 Creating holorepository-core_holostorage-accessor_1                          ... done
 ```
 
+In cases where only selective services are required, they can be started by specifying the service name
```
nielsboecker: This seems a bit redundant with line 177 - 180 now, don't you think?

nielsboecker (approved):
One minor comment. Generally good to merge

README.md
boonwj: oh yes true, I missed that part.

PR #64, opened 2019-08-21
Ui/fake smart auth

fanbomeng97:
Implemented the fake SMART auth process: 

- Now you can select the practitioner form a dropdown, and the back-end server will request related information with the selected practitioner ID

- DeviceConnectorPage will display the PIN for cureent user

- Move FHIR server url to .env

### DIFF
--- a/HoloRepositoryUI/client/src/components/app/DeviceConnectorPage.tsx
+++ b/HoloRepositoryUI/client/src/components/app/DeviceConnectorPage.tsx
@@ -1,8 +1,9 @@
 import React, { Component } from "react";
 import { RouteComponentProps } from "@reach/router";
 import ContentContainer from "./core/ContentContainer";
+import { PropsWithContext, withAppContext } from "../shared/AppState";
 
-class DeviceConnectorPage extends Component<RouteComponentProps> {
+class DeviceConnectorPage extends Component<RouteComponentProps & PropsWithContext> {
   render() {
     return (
       <ContentContainer
@@ -13,10 +14,10 @@ class DeviceConnectorPage extends Component<RouteComponentProps> {
           Use the code on this page to link your HoloLens application with your SMART on FHIR login.`
         ]}
       >
-        <div>This feature is not available yet.</div>
+        <div>Your authentication PIN: <b>{this.props.context!.pin}</b></div>
       </ContentContainer>
     );
   }
 }
 
-export default DeviceConnectorPage;
+export default withAppContext(DeviceConnectorPage);
--- a/HoloRepositoryUI/client/src/components/public/LandingPage.tsx
+++ b/HoloRepositoryUI/client/src/components/public/LandingPage.tsx
@@ -12,6 +12,7 @@ import {
 import { Layout } from "antd";
 import appLogo from "../../assets/logo/2x/logo_and_font@2x.png";
 import MainFooter from "../shared/MainFooter";
+import { PropsWithContext, withAppContext } from "../shared/AppState";
 
 const options: IDropdownOption[] = [
   { key: "b0016666-1924-455d-8b16-92c631fa5207", text: "Isaac Upton" },
@@ -26,7 +27,9 @@ const imageProps: IImageProps = {
   maximizeFrame: true
 };
 
-class LandingPage extends Component<RouteComponentProps> {
+class LandingPage extends Component<RouteComponentProps & PropsWithContext> {
+  selectedItem: string | number = "";
+
   backgroundStyle = {
     background: "rgba(1,1,1,0.2)"
   };
@@ -65,6 +68,7 @@ class LandingPage extends Component<RouteComponentProps> {
               placeholder="Select an practitioner"
               label="Practitioner:"
               options={options}
+              onChanged={item => this._onChange(item.key)}
             />
             <TextField label="Password:" type="password" style={{ width: "300" }} />
             <div
@@ -75,7 +79,12 @@ class LandingPage extends Component<RouteComponentProps> {
               }}
             >
               <Link to="/app/patients">
-                <PrimaryButton data-automation-id="login" text="Login" allowDisabledFocus={true} />
+                <PrimaryButton
+                  data-automation-id="login"
+                  text="Login"
+                  allowDisabledFocus={true}
+                  onClick={this._selectPractitioner}
+                />
               </Link>
             </div>
           </div>
@@ -85,6 +94,14 @@ class LandingPage extends Component<RouteComponentProps> {
       </Layout>
     );
   }
+
+  private _selectPractitioner = () => {
+    this.props.context!.handleLogin(this.selectedItem);
+  };
+
+  private _onChange = (key: string | number): void => {
+    this.selectedItem = key;
+  };
 }
 
-export default LandingPage;
+export default withAppContext(LandingPage);
--- a/HoloRepositoryUI/client/src/components/public/LandingPage.tsx
+++ b/HoloRepositoryUI/client/src/components/public/LandingPage.tsx
@@ -13,12 +13,12 @@ import { Layout } from "antd";
 import appLogo from "../../assets/logo/2x/logo_and_font@2x.png";
 import MainFooter from "../shared/MainFooter";
 import { PropsWithContext, withAppContext } from "../shared/AppState";
+import { string } from "prop-types";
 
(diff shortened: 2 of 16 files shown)


fanbomeng97:
Seems like the server test is broken, i have checked it, it caused by moving fhir url to ` .env`, but so far no idea how to fix it ... 
Going to modify the hololens app, pls have a look when you review the pr

nielsboecker:
Cool, I'll take a look soon.
@fanbomeng97 as for the missing variable, maybe [this](https://github.com/vuejs/vue-test-utils/issues/193#issuecomment-438437626) can help?

Edit: Fixed. It was due to an import from the `fhirClient` file. So the top level code of that script ran (even before any jest setup code). So I changed that.

HoloRepositoryUI/server/src/common/clients/fhirClient.ts
```diff
@@ -2,13 +2,14 @@ import Client from "fhir-kit-client";
 import t from "io-ts";
 import { decode, isDecodeError } from "io-ts-promise";
 import { getAdapterFunction } from "../adapters/fhirAdapter";
-import { FHIR_SERVER_BASE_URL } from "../../config";
 import { R4 } from "@ahryman40k/ts-fhir-types";
 import { IImagingStudy, IPatient, IPractitioner } from "../../../../types";
 import logger from "../logger";
 
+const { FHIR_SERVER_BASE_URL: fhirUrl } = process.env;
```
nielsboecker: I'd argue that there is no need for the alias, could just be
```
const { FHIR_SERVER_BASE_URL } = process.env;
```

HoloRepositoryUI/server/.env.local
```diff
@@ -9,3 +9,6 @@ HOLOPIPELINES_HOST=http://localhost
 HOLOPIPELINES_PORT=3100
 HOLOSTORAGE_ACCESSOR_HOST=http://localhost
 HOLOSTORAGE_ACCESSOR_PORT=3200
+
+# Endpoints of FHIR server
+FHIR_SERVER_BASE_URL=https://holoehr-demo.azurewebsites.net
```
nielsboecker, line 14: This is a really small issue, but I notice that you sometimes delete the newline at the end of the file. It is actually good practice to have one empty line at the end. Some code formatters will even add it. So no need to delete it :)

HoloRepositoryUI/client/src/components/app/DeviceConnectorPage.tsx
```diff
@@ -13,10 +14,10 @@ class DeviceConnectorPage extends Component<RouteComponentProps> {
...
       >
-        <div>This feature is not available yet.</div>
+        <div>Your authentication PIN: <b>{this.props.context!.pin}</b></div>
       </ContentContainer>
     );
   }
 }
 
-export default DeviceConnectorPage;
+export default withAppContext(DeviceConnectorPage);
```
nielsboecker: I don't understand why `AppComponents.smoke.test` didn't fail. I was under the impression that you'd have to `mountWithContextProvider` when you use `AppContext` in the component. :/

Do you understand what's going on?

HoloRepositoryUI/client/src/components/app/DeviceConnectorPage.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { Component } from "react";
 import { RouteComponentProps } from "@reach/router";
 import ContentContainer from "./core/ContentContainer";
+import { PropsWithContext, withAppContext } from "../shared/AppState";
 
-class DeviceConnectorPage extends Component<RouteComponentProps> {
+class DeviceConnectorPage extends Component<RouteComponentProps & PropsWithContext> {
```
nielsboecker: Actually the component name isn't great. It is still from when we had a different plan for this. Maybe we can rename it. How about ... `HoloLensRemoteLoginPage` or something better

HoloRepositoryUI/client/src/components/public/LandingPage.tsx
```diff
@@ -1,15 +1,114 @@
...
+            display: "flex",
+            justifyContent: "center"
+          }}
+        >
+          <div style={{ ...this.LoginContainer }}>
+            <div style={{ ...this.loginLogo }}>
+              <Image {...(imageProps as any)} alt="app_logo" />
+            </div>
+            <Dropdown
+              placeholder="Select an practitioner"
```
nielsboecker: Maybe more along the lines of "Select your account"

HoloRepositoryUI/client/src/components/public/LandingPage.tsx
```diff
@@ -1,15 +1,114 @@
...
   }
+
+  private _selectPractitioner = () => {
+    if (this.selectedItem === ""){
+      this.selectedItem = options[0].key;
+    }
+
+    if (typeof this.selectedItem !== "number") {
+      let id = this.selectedItem.substring(0, this.selectedItem.indexOf("/"));
+      let pin = this.selectedItem.substring(this.selectedItem.indexOf("/") + 1);
```
nielsboecker: you can use `const` instead of `let`.

HoloRepositoryUI/client/src/components/shared/App.tsx
```diff
@@ -69,7 +69,9 @@ class App extends Component<any, IAppState> {
       for (const pid in combinedResult) {
         const studies = combinedResult[pid];
         const patient = patients[pid];
-        patient.imagingStudies = studies;
+        if (patient !== undefined) {
```
nielsboecker: Could just be `if (patient)`

Also, did you add this because something went wrong? Seems to me like this should actually never happen because you just query for the IDs you have actual patients for.

nielsboecker (commented):
The login page is beautiful! Just some minor comments. 

(Plus we have to fix the failing build)

nielsboecker:
Oh, another idea! Maybe you can also show the HoloRepository logo above the login screen of the HoloLens app, to make it a more consistent experience.

HoloRepositoryUI/client/src/components/public/LandingPage.tsx
```diff
@@ -1,15 +1,114 @@
...
+  { key: "a102/03827", text: "Jonah Schroeder" }
+];
+
+const imageProps: IImageProps = {
+  src: appLogo,
+  imageFit: ImageFit.cover,
+  maximizeFrame: true
+};
+
+class LandingPage extends Component<RouteComponentProps & PropsWithContext> {
```
nielsboecker: Maybe we should rename it to LoginPage now

HoloRepositoryUI/client/src/components/public/LandingPage.tsx
```diff
@@ -1,15 +1,114 @@
...
+  IImageProps,
+  ImageFit,
+  Image
+} from "office-ui-fabric-react";
+import { Layout } from "antd";
+import appLogo from "../../assets/logo/2x/logo_and_font@2x.png";
+import MainFooter from "../shared/MainFooter";
+import { PropsWithContext, withAppContext } from "../shared/AppState";
+
+const options: IDropdownOption[] = [
```
nielsboecker: Could you add a comment that these are hard-coded because the authentication is just a proof-of-concept for now?

HoloRepositoryUI/client/src/components/shared/App.tsx
fanbomeng97: When the practitioner id is wrong, this code will crash because patient is undefined, but now we use the hard-coded practitioner list, it won't happens anymore

HoloRepositoryUI/client/src/components/app/DeviceConnectorPage.tsx
fanbomeng97: No idea :joy:

nielsboecker: Oh I remember. If you do a "shallow render" and you have a HOC (Higher Order Component), it will just mount the layer of this wrapper component. The actual component, being a child of that wrapper, will not get mounted, so it's not really a test. Therefore it would be better if you could add that to the test plase :)

nielsboecker (approved):
Great work Fanbo. After adding the one change in the test suite, this is good to merge!

PR #60, opened 2019-08-19
HoloRepository Demo Setup

boonwj:
This PR is for the creation of a HoloRepository Demo setup.

It includes the steps, demo data and changes to the tools necessary to set it up. The following provides an overall view of what is included in.

- Demo data for EHR FHIR server
- Demo data for HoloStorage FHIR server
- Demo data for HoloStorage Blob service
- Additional feature for `az_blob_tool` to allow upload of a folder of files
- Additional feature for `fhir_api_tool` to upload files containing individual FHIR resource
- Change `Holostorage Accessor` to retain file extension when storing data in blob storage 
- Change Postman query collection to be related to demo data 

TODO: UI end is not tested with the demo data. @fanbomeng97  or @nbckr, to look into it

PR #65, opened 2019-08-21
Misc/load test

boonwj:
This PR brings in load testing for the UI and Accessor components.

Testing is done using `https://locust.io/` and the tests works with the assumption that the backend services (ehr and holostorage) are loaded with FHIR data from `holorepo_demo' (#60)

The tests currently only cover the GET/PUT endpoints. The POST endpoint will generate many unique ids which will be hard to clean up.

### DIFF
--- a/Misc/tools/locust_load_test/README.md
+++ b/Misc/tools/locust_load_test/README.md
@@ -8,12 +8,14 @@ It tests the endpoints of the `UI` and `Accessor` components of the HoloReposito
 - HoloRepository Azure FHIR Service Infra (EHR and HoloStorage)
 
 ## Testing
+> Note: Testing is done with the assumption that the service contains data from the `holorepo_demo`. View `Misc/holorepo_demo` for more details.
+
 To execute locust load test on the accessor or UI.
 
 First install LocustIO with `pipenv install --dev`
 
-Then, depending on what you are testing, run the following components.
-- Testing UI: Accessor and UI
+Then, depending on what you are testing, the following components must be running.
+- Testing UI: UI and Accessor
 - Testing Accessor: Accessor
 
 Next, start the locust test service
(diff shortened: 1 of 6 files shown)


Misc/tools/locust_load_test/README.md
```diff
@@ -0,0 +1,33 @@
+# LocustIO Load Test
```
nielsboecker, line 1: This README could be more specific about what UI means (client? server? both?)

Misc/tools/locust_load_test/locustfile-ui.py
```diff
@@ -0,0 +1,45 @@
+from locust import HttpLocust, TaskSet, task
+
+
+class UserBehavior(TaskSet):
+    @task
+    def getPatients(self):
+        self.client.get("/api/v1/patients")
```
nielsboecker, line 7: ~~Maybe api prefix could be extracted to a variable~~ Actually i kinda like it like it is

Misc/tools/locust_load_test/locustfile-ui.py
```diff
@@ -0,0 +1,45 @@
...
+    def getPractitioner(self):
+        self.client.get("/api/v1/practitioners/a100")
+
+    @task
+    def getImagingStudies(self):
+        self.client.get("/api/v1/imagingStudies")
+
+    @task
+    def getImagingStudiesOfPatients(self):
+        self.client.get("/api/v1/imagingStudies?pids=p100,p102")
```
nielsboecker, line 27: 😸

nielsboecker (approved):
Looks great! Would you also like to quickly summarise how it performs in OneNote, and/or write a section in the report on this?

Misc/tools/locust_load_test/README.md
boonwj: I am really just hitting the server endpoints. So I will just change it to UI-Server

Issue #62, opened 2019-08-20
PIPELINE: 3D models in bone_segmentation pipeline are weird


nielsboecker:
![image](https://user-images.githubusercontent.com/11090412/63391154-6a310e80-c3a9-11e9-9dc7-af84fe871df8.png)

This might be an issue introduced by my refactoring #45.

I don't think it's related to #57 as this is DICOM only, no NIfTI involved.

nielsboecker:
I suspect the problem somewhere around `dicom_file.py` line 110. Because I now fix the orientation _before_ doing the normalising a.k.a. resampling. @UdomkarnBoonyaprasert I think you have a better understanding of this. When you have time, maybe you can branch off the refactoring branch and help with this issue.

PR #66, opened 2019-08-21
Pipeline: fix scale factor

UdomkarnBoonyaprasert:


nielsboecker (approved):
Thanks, I verified that it works:

image

(although not sure if we should consider maybe another HU threshold...)

nielsboecker:
Resolves #62

PR #70, opened 2019-08-22
MISC: Fix locust ui hologram test endpoints

boonwj:
- Remove unsupported endpoint /holograms
- Fix query argument to pids instead of hids


nielsboecker (approved):
👍

PR #71, opened 2019-08-22
PIPELINE: Fix config bug

nielsboecker:
Boolean and numerical values were parsed as strings, d'oh


nbckr:
Note: There are some other places in the code where I read the numeric variables and cast them, which is now redundant. I will fix this in the othher branch I'm working on atm. I excluded it here as I made some other changes in these files and wanted to keep this PR here small.

HoloPipelines/jobs/jobs_state.py
```diff
@@ -34,32 +41,33 @@
 )
 
 
-def update_job_state(job_id: str, new_state: str, logger=logging):
+def update_job_state(job_id: str, new_state: str, logger=logging, new: bool = False):
     """
-    Updates the global dictionary that keeps track of all jobs. Note that new_state
+    Updates the state for aa job by updating the job.state file. Note that new_state
```
boonwj, line 46: typo "a"

HoloPipelines/jobs/jobs_io.py
```diff
@@ -10,55 +10,106 @@
...
+def get_all_job_subdirectory_names() -> List[str]:
+    return [dir.name for dir in os.scandir(jobs_root) if dir.is_dir()]
+
+
+def read_state_file_for_job(jod_id: str) -> Tuple[str, float]:
+    """
+    Reads state file and returns state and time in seconds since last modification.
+    """
+    state_file_path = get_state_file_path_for_job(jod_id)
+    with open(state_file_path, "r") as state_file:
```
boonwj, line 96: There needs to be error handling here, as querying a non-existent job id will throw an exception

HoloPipelines/jobs/jobs_io.py
```diff
@@ -80,16 +131,28 @@ def get_logger_for_job(job_id: str):
...
+    """
+    Returns the path to a log file. Looks for current jobs first. If nothing found,
+    looks for finished jobs. If still not found, returns None.
+    """
+    path = get_directory_path_for_job(job_id)
+    if os.path.isdir(path):
+        return f"{path}/job.log"
+    elif os.path.isdir(path.replace(jobs_root, finished_jobs_root)):
+        return f"{path.replace(jobs_root, finished_jobs_root)}/job.log"
+
```
boonwj, line 144: An explicit `return None` will be idiomatic given that the function returns different values and None is one of them

HoloPipelines/jobs/jobs_io.py
```diff
@@ -10,55 +10,106 @@
...
+    return os.path.exists(get_state_file_path_for_job(job_id))
+
+
+def write_state_file_for_job(jod_id: str, state: str) -> None:
+    state_file_path = get_state_file_path_for_job(jod_id)
+    with open(state_file_path, "w") as state_file:
+        state_file.write(state)
 
 
 def get_logger_for_job(job_id: str):
```
boonwj, line 115: Optional: type hints for return value?

HoloPipelines/jobs/jobs_state.py
```diff
@@ -5,17 +5,24 @@
...
+    get_all_job_subdirectory_names,
+    read_state_file_for_job,
+    remove_temporary_data_for_job,
+    remove_log_file_for_job,
+    move_job_to_finished_jobs_directory,
+    write_state_file_for_job,
+    state_file_for_job_exists,
+)
 
 JobState = Enum(
```
boonwj, line 27: `JobState` does not feel like a proper python variable naming convention for global variables. The naming convention should be similar to python functions.

https://www.python.org/dev/peps/pep-0008/#global-variable-names

boonwj (approved):
No reason to reject the config bug fixes as this is part of the major refactoring changes. 

In addition, this PR also introduces changes to the way job states are tracked and managed.

It's best to merge it with the major-refactor branch and be evaluated as a whole.

HoloPipelines/jobs/jobs_io.py
nielsboecker: Agree, did it in the other branch

HoloPipelines/jobs/jobs_state.py
nielsboecker: Hm but it is an Enum right? See for instance how they use it in the docs: https://docs.python.org/3/library/enum.html#functional-api

PR #67, opened 2019-08-21
PIPELINE: It's the end of HoloPipelines as we know it

nielsboecker:
This PR reflects the gigantic #45 refactoring that I did to the HoloPipelines. I did a whole bunch of things and nothing is like it used to be. Hallelujah!

Some major points:
* complete re-architecture of all components
* implement several features that were kinda half-baked before
* fix numerous issues and problems
* implement dynamic pipeline loading
* do some basic error handling
* use `multiprocessing` instead of `subprocess`
* job state in-memory instead of POSTing to self
* job-specific data kept in job's own directory
* automatic garbage collection
* add logging to file per job and serve that at a new endpoint
* add type hints and docstrings
* add documentation
* dockerise application
* add Postman collection
* identify further problems and inconsistencies, for which I left TODOs or opened issues

> **Disclaimer 1**: There are still many TODOs in the code. The test coverage is now zero. This must be worked on within following issues and PRs.

> **Disclaimer 2**: I have mostly worked with the `bone_segmentation` pipeline and can confirm that it works end-too-end. The other ones @ansonwong9695 is still working at confirming nothing broke.

PR #73, opened 2019-08-22
PIPELINE: gunicorn entrypoint

nielsboecker:
When you run the server through gunicorn, the `__main__` doesn't run, so to fix that I followed the SO recommendation and put the initialisation to the top level (so it happens at import time)

However I am not sure if this is the right way to go. I haven't really verified it running in the gunicorn Docker container, as my computer is terribly slow. I think it may be a problem if/when we start gunicorn with multiple workers.

If we keep gunicorn workers at 1, as we have now, and start multiple processes in our own code, it should be fine? I noticed that when I just run Flask without gunicorn in development mode, it uses two processes, whereas when i run it with FLASK_ENV=production, it uses only 1. When I had two, they both tried to do the garbage collection file deletion. I added some error handling in #72 to solve that problem.


nielsboecker:
@boonwj Not having this fix could also be related with the errors you mention in #67, as the initialisation from main method is not running when you use gunicorn.

PR #76, opened 2019-08-25
UI/BE: Use variables from process.env

nielsboecker:
That was dumb 🤕 Some variables, even though they were already prepared in `.env`, were not actually accessed through the environment variable.


boonwj (approved):
Thanks for the quick fix. This should fix the deployment

Issue #56, opened 2019-08-17
PIPELINE: Center cropping is faulty


nielsboecker:
While doing the refactoring for #45 I realised that the centerCrop method is currently useless, as it is only conditionally called. The condition is always false, as it is checked right after downscaling is performed.

Also, @UdomkarnBoonyaprasert where did the whole logic go that made sure the image size is divisable by 8?

PR #77, opened 2019-08-26
Pipeline: Update NumPy array cropping condition

UdomkarnBoonyaprasert:
Cropping is now done if any of the dimensions is not divisible by 8. This is to address issue #56

### DIFF
--- a/HoloPipelines/core/services/np_image_manipulation.py
+++ b/HoloPipelines/core/services/np_image_manipulation.py
@@ -10,30 +10,46 @@ import scipy.ndimage
 from config import INPUT_RESOLUTION_MAX
 
 
-# FIXME: This is broken and needs to be fixed @Udomkarn
+def crop_around_centre(
+    image: np.ndarray,
+    new_x_dimension_length,
+    new_y_dimension_length,
+    new_z_dimension_length,
+) -> np.ndarray:
+    """
+    Crop the 3D NumPy array around the centre to a new dimension given as arguments.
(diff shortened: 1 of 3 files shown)


HoloPipelines/core/services/np_image_manipulation.py
```diff
@@ -51,11 +51,11 @@ def downscale_and_conditionally_crop(
         x = image.shape[0]
```
nielsboecker: The FIXME comment in line 13 should be removed if you fixed it

nielsboecker: crop_around_centre should have a doc string and the arguments should have better names

UdomkarnBoonyaprasert: FIXME comment removed and arguments naming updated

nielsboecker:
@UdomkarnBoonyaprasert Code looks great now 👍. How about adding a test before we close the issue? To assert the different scenarios work:
```
np_image_manipulation_test.py
    test_crop_around_centre
    test_downscale_and_conditionally_crop_with_only_downscale
    test_downscale_and_conditionally_crop_with_downscale_and_crop
    test_downscale_and_conditionally_crop_with_only_crop
```

nielsboecker (approved):
Thanks @UdomkarnBoonyaprasert, nice testing!
FYI please take a look at this small change I made: 915b319
I will merge now, and this resolves #56

PR #79, opened 2019-08-26
Pipeline: test dicom_file inside adapter

ansonwong9695:
Add test for the dicom_file.py


nielsboecker:
Nice work @ansonwong9695! Please take a look at the small changes I made. I will merge this now. I like this kind of small PRs btw!

PR #82, opened 2019-08-27
Pipeline: Simplify wrapper 

ansonwong9695:
- add ```Simplify``` binary to ```third_party```
- add its wrapper in ```wrappers```
- add ```SIMPLIFICATION_RATIO``` in env as default value
- add simplify calls to all pipelines


HoloPipelines/core/wrappers/simplify.py
```diff
@@ -0,0 +1,29 @@
+import logging
+import os
+import subprocess
+
+from config import SIMPLIFICATION_RATIO
+
+current_script_directory = (
```
nielsboecker: replace with `simplify_binary_path = "./core/third_party/fast_quadric_mesh_simplification/simplify"`

nielsboecker: Can you move the binary to that subdirectory and create a small README in the directory too please, stating where it is taken from. Like [>this<](https://github.com/nbckr/HoloRepository-Core/blob/PIPELINE/major-refactoring/HoloPipelines/core/third_party/ct_lung_segmentation/README.md).

HoloPipelines/core/wrappers/simplify.py
```diff
@@ -0,0 +1,29 @@
...
+    os.path.dirname(os.path.realpath(__file__)) + "/../third_party/"
+)
+
+
+def call_simplify(
+    obj_input_path: str,
+    obj_output_path: str,
+    simplification_ratio: float = SIMPLIFICATION_RATIO,
+):
+    # NOTE: link to simplify repository https://github.com/sp4cerat/Fast-Quadric-Mesh-Simplification
```
nielsboecker: Remove from here

HoloPipelines/core/wrappers/simplify.py
```diff
@@ -0,0 +1,29 @@
...
+    simplify_command = [
+        "./simplify",
```
nielsboecker: Replace with `simplify_binary_path`

HoloPipelines/core/wrappers/simplify.py
```diff
@@ -0,0 +1,29 @@
...
+        obj_input_path,
+        obj_output_path,
+        str(simplification_ratio),
+    ]
+
+    completed_process = subprocess.run(simplify_command, cwd=current_script_directory)
```
nielsboecker: Remove the `cwd`

nielsboecker (commented):
Code looks pretty good! Please look at my comments.

Also can you please add a simple test? It's sufficient to just call the method and verify that a file exists in the output path afterwards. You may want to find a very small obj to include in the `utils` directory as input for that purpose.

nielsboecker:
Btw @ansonwong9695, @UdomkarnBoonyaprasert I merged this locally with the tests branch and the 3 pipeline tests fail. I think it's because the path to the `temp.obj` file is broken due to the way you use `cwd` (=> it will look for the temp file as subdirectory of the cwd path, which is of course wrong). So I suspect that once you follow my PR suggestions, it will also fix that bug.

nielsboecker (changes requested):
Paths are broken (again)

nielsboecker:
**TODO**: Once my comments are sorted out, change PR target to `dev`

HoloPipelines/core/wrappers/simplify.py
UdomkarnBoonyaprasert: moved and ```README.md``` added

UdomkarnBoonyaprasert:
I added a sample obj file (```tests/utils/teddy.obj```) for testing. It's around 90 KB, would this be ok?

nielsboecker:
Yes that sounds like a good idea. But currently, the test fails, maybe due to absolute vs relative path to test data.

nielsboecker (approved):
Thanks @UdomkarnBoonyaprasert I have made some refactorings and I am happy with this PR now. @boonwj Going to leave it up to you to do the final merge, as you are currently deploying the system.

I have verified that the bone + lung pipeline are still functional locally. Haven't tried dockerised versions nor the abdominal pipeline, but it should be fine, as it is just the additional step with the simplify wrapper.

boonwj:
The different workflows do appear to work in docker. There's just a new configuration field (SIMPLIFICATION_RATIO) that I need to take care of when I deploy it to kubes.

Also, there does not seem to be any documentation on the individual configuration fields. Pipeline's readme and the configuration section only mentioned that .env is the configuration file. 

If there's time, it will be nice to document the individual fields.

PR #83, opened 2019-08-27
PIPELINE: Use HTTP client in receive_input task

nielsboecker:
`http_client` was extracted in a PR. I then realised that I introduced duplicate code, so refactored it.

PS: The "shared tasks" are currently really only redirecting to other components. I think it's okay to leave it as is, and was thinking that in the future they can be used to make the pipelines even more generic (always just plug together a couple of tasks, without knowing if it's a wrapper, client, service...) but if it bothers you in its current state, they could be removed and replaced with the direct calls in the pipelines.

### DIFF
--- a/HoloPipelines/core/clients/http.py
+++ b/HoloPipelines/core/clients/http.py
@@ -2,7 +2,7 @@
 This module contains provides a generic HTTP client with helper methods for
 fetching data from the web.
 """
-
+import io
 import logging
 from zipfile import ZipFile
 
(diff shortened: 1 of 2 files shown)


boonwj:
This refactor breaks the download process.

```
holopipelines-core_1                             | 2019-09-08 10:56:51 | root | INFO  | Download of 'https://holoblob.blob.core.windows.net/mock-pacs/left-scfe-pelvis-bone.zip' was successful
holopipelines-core_1                             | 2019-09-08 10:56:51 | root | WARNING | [ERROR] An error occurred and caused the job to fail: type object 'bytes' has no attribute 'seek'
```

boonwj (changes requested):
This change breaks the main pipelines

nielsboecker:
@boonwj Try again, please. I fixed what appears to be a rather obvious bug, and verified with a simple Postman request on my machine that the pipeline download and extraction works.

boonwj (approved):
Works smoothly without errors.

PR #69, opened 2019-08-22
Pipeline: fix nifti normalisation

UdomkarnBoonyaprasert:

PR #84, opened 2019-08-28
PIPELINE: Fix lung segmentation

nielsboecker:
🐛 @UdomkarnBoonyaprasert updated function signatures in  #69, but the parts of the lung pipeline that access this module were not updated. So everything crashed:

image


boonwj (approved):
The pipeline worked with this.

PR #85, opened 2019-08-28
Ui/fix technical debt

fanbomeng97:
Fix some bug and design issues:

- width of main content should be consistent

- bug with header height

- bug with footer/app height

- redirect to login  page when going to random page without login


nielsboecker (commented):
* I <3 that the header finally has one height! And the main content has same width
* the way you changed href caused an error in the console cause `render` didn't return anything; I changed that, and also updated the logout method (edit: you fixed it too in the meantime, but my solution seems still cleaner, using the provided component by reach-router)

nielsboecker (approved):
💯

PR #86, opened 2019-08-28
Pipeline/callback fix

boonwj:
Quick fix for the callback function

### DIFF
--- a/HoloPipelines/jobs/jobs_controller.py
+++ b/HoloPipelines/jobs/jobs_controller.py
@@ -6,7 +6,7 @@ new jobs, does various checks, and performs some high-level error handling.
 import logging
 import uuid
 from multiprocessing import Pool
-from typing import Tuple
+from typing import Tuple, Optional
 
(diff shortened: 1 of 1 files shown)


HoloPipelines/jobs/jobs_controller.py
```diff
@@ -47,10 +47,13 @@ def check_job_request_validity(job_request: dict) -> Tuple[bool, str]:
     return True, ""
 
 
-def job_success_callback() -> None:
+def job_success_callback(result: Optional[any]) -> None:
```
nielsboecker: I think in the other files I used `Any` (capital A), which I believe also neds to be imported from `typing`. But I don't know about `any`; if this also works, may even be the nicer solution.

nielsboecker (approved):
💡

HoloPipelines/jobs/jobs_controller.py
boonwj: oh ya that's a built-in method not a type. Fixed accordingly.

PR #75, opened 2019-08-24
PIPELINE: Add test fixtures and refurbish test suite

nielsboecker:
This PR adds a whole couple of tests for the new HoloPipelines architecture. It also includes reusable fixtures that are shared for many tests and can be used for future tests.

It's mostly just these tests, so having the build pipeline succeed should be a good criteria to accept the PR. There are also some minor changes in the business logic code, that were made when problems were discovered by writing tests.

PR #87, opened 2019-08-28
PIPELINE: Update .coveragerc

nielsboecker:
This is a quick fix, extracting one change from #75 so that testing on the build server can be reenabled immediately.


nielsboecker:
I turned the test step back on, it passes: [Build](https://dev.azure.com/MSGOSHHOLO/HoloRepository/_build/results?buildId=484)

nielsboecker:
Faulty git history, re-did it at #88

PR #88, opened 2019-08-28
PIPELINE: Update pytest config and add test input file

nielsboecker:
This is a quick fix, extracting one change from #75 so that testing on the build server can be reenabled immediately.

I turned the test step back on, it passes: [Build](https://dev.azure.com/MSGOSHHOLO/HoloRepository/_build/results?buildId=484)


boonwj (approved):
ok looks valid

PR #89, opened 2019-08-28
PIPELINE: add testing for job controller and job state

ansonwong9695:


nielsboecker (commented):
Good work @ansonwong9695, thanks. I made some refactorings, please take a look.

I also noticed a strange behaviour of a test unrelated to this PR, the lung_segmentation. When run alone, all is well:
image

When running as part of a complete run of all tests however, it seems to weirdly run twice at the same time? And it seems to try to talk to the Accessor, even though that part should me mocked out? Let alone that thhe error comes from job_controller, which should not even be involved in this test?!?
image

The test still passes. My feeling: Something going wrong due to imports and initialisation of module variables somewhere. Due to the time constraints I tend toward still merging it and living with it. @boonwj  You agree?

nielsboecker:
**Update**: Have been analysing some more and found the problem. It was actually that @ansonwong9695 you started a real job here:
```python
def test_start_new_job():
    result_success, result_job_response = start_new_job(sample_job)
```
This job continued running, and in the course of subsequent tests, caused unexpected behaviour (due to the Accessor call not being mocked). The reason that only `lung` pipeline was affected is that the tests run in alphabetical order, and the other pipelines run before the `jobs` stuff. 💩 

I will add a fix.

nielsboecker (approved):
With my fix, this should be fine. @ansonwong9695 For your reference, please take a look at the changes I made.

Issue #48, opened 2019-08-14
PIPELINE: Abdominal pipeline has 8 outputs instead of 1


nielsboecker:
# Description
Currently, the abdominal pipeline takes the segmentation output and splits it into 8 parts. There are some problems here:
* the one and only reason to do this is colouring, and currently it looks like there is no easy solution for that according to @UdomkarnBoonyaprasert 
  * => probably out-of-scope for the two weeks we have left!
* performance-wise, this means that the time needed for steps afterwards is multiplied by 8: 
  * Ray marching algorithm (converting numpty to obj, and then obj to glb)
  * Simplification and compression
  * this means, instead of approximately `2 + 2 sec`, the post-processing will now take around `8 * 2 + 8 * 2 sec`
* There is currently no component that merges the 8 outputs back into 1 file
  * The final output has to be one file, so if you split them, you also have to merge them
  * We are currently not certain if this is an easy or laborous task

## Suggested solution
Given our time constraints, it might be best to not even split the output of the NN?

nielsboecker:
**Update**
Discussed with @boonwj.
* the performance drawback is negligible
* probably we have two options to go about this
  * 1: keep the 8 different organs and create a hologram for each
    * This will require to have the appropriate names for each organ, not just "organ1", "organ2", which could be done by @UdomkarnBoonyaprasert 
    * pro: conceptually interesting, good in report. con: looks very ugly.
  * 2: keep the 8 different organs, do the colouring, merge them and post 1 hologram.
    * this will require some more research whether colouring is possible
    * pro: may look cool. con: technically challenging, maybe infeasible.

The decision is currently postponed. Next steps would be for someone (maybe @fanbomeng97 or myself) to explore the gltf colouring more. Then we decide.

nielsboecker:
We should decide what to do here soon.

nielsboecker:
This problem is not solved yet, and we are running out of time. Therefore, we are going to stick with just the segmented output as 1 model, and with no added colours.

@UdomkarnBoonyaprasert / @ansonwong9695 Can you please add a comment with explanation and deprecation note to the script that splits apart organs.

nielsboecker:
Also, the output when I ran this pipeline looks weird:
![image](https://user-images.githubusercontent.com/11090412/63939349-0076cb80-ca5f-11e9-97e3-a57949d5c403.png)

[...]

PR #91, opened 2019-08-29
Pipeline: Abdomen resolution fix

ansonwong9695:
This PR is to address issues from #48
- added Y axis flipping function exclusive to abdomen pipeline
- added dynamic config update for the resolution of the in dense v net model
- updated ```requirements.txt``` in dense v net model
- updated ```dockerfile``` in dense v net model
- added comment with explanation and deprecation note to the script that splits apart organs in ```split_to_seperate_organs.py```

### DIFF
--- a/HoloPipelines/core/pipelines/abdominal_organs_segmentation.py
+++ b/HoloPipelines/core/pipelines/abdominal_organs_segmentation.py
@@ -13,7 +13,10 @@ import os
 import sys
 
 from config import MODEL_ABDOMINAL_SEGMENTATION_HOST, MODEL_ABDOMINAL_SEGMENTATION_PORT
-from core.adapters.dicom_file import read_dicom_as_np_ndarray_and_normalise
+from core.adapters.dicom_file import (
+    read_dicom_as_np_ndarray_and_normalise,
+    flip_numpy_array_dimensions_y_only,
+)
 from core.adapters.glb_file import convert_obj_to_glb_and_write
 from core.adapters.nifti_file import (
     convert_dicom_np_ndarray_to_nifti_image,
@@ -49,6 +52,7 @@ def run(job_id: str, input_endpoint: str, medical_data: dict) -> None:
 
     update_job_state(job_id, JobState.READING_INPUT.name, logger)
     dicom_image_array = read_dicom_as_np_ndarray_and_normalise(dicom_directory_path)
+    dicom_image_array = flip_numpy_array_dimensions_y_only(dicom_image_array)
     crop_dicom_image_array = downscale_and_conditionally_crop(dicom_image_array)
 
     update_job_state(job_id, JobState.PREPROCESSING.name, logger)
--- a/HoloPipelines/core/tasks/abdominal_organs_segmentation/split_to_separate_organs.py
+++ b/HoloPipelines/core/tasks/abdominal_organs_segmentation/split_to_separate_organs.py
@@ -1,14 +1,15 @@
 """
 This module contains the functionality of the abdominal_organs_segmentation pipeline
 of splitting the combined result into multiple sub-results.
-"""
 
-# TODO: Depending on how we decide for this pipeline, this component may get removed.
-# Otherwise, separate organs need to be merged again! (can also be in this task, which
-# should then be renamed though)
-# TODO: If we keep this, refactor and verify
-# Refactoring should include only doing the OBJs in here, and transforming to GLB on
-# pipeline level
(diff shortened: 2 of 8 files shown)


HoloPipelines/core/tasks/abdominal_organs_segmentation/split_to_separate_organs.py
```diff
@@ -1,14 +1,15 @@
...
-"""
 
-# TODO: Depending on how we decide for this pipeline, this component may get removed.
-# Otherwise, separate organs need to be merged again! (can also be in this task, which
-# should then be renamed though)
-# TODO: If we keep this, refactor and verify
-# Refactoring should include only doing the OBJs in here, and transforming to GLB on
-# pipeline level
+Currently this module is not in used. The current implementation of
+this module is unable to provide an appropriate name to each output
```
nielsboecker, line 6: Hm actually we can provide a name that's not the problem. Problem is we can't assign colours, and can't merge them. But this can be made shorter:

"Currently, this component is unused, as the abdominal_organs_segmentation pipeline yields only one output. This might be integrated later. Note that a further step will need to be implemented, to either merge the 8 models back into 1, or make the pipeline yield 8 outputs for one job."

HoloPipelines/core/pipelines/abdominal_organs_segmentation.py
```diff
@@ -49,6 +52,7 @@ def run(job_id: str, input_endpoint: str, medical_data: dict) -> None:
 
     update_job_state(job_id, JobState.READING_INPUT.name, logger)
     dicom_image_array = read_dicom_as_np_ndarray_and_normalise(dicom_directory_path)
+    dicom_image_array = flip_numpy_array_dimensions_y_only(dicom_image_array)
```
nielsboecker, line 55: But _why_ do you flip it?

Please add a comment before this line.

HoloPipelines/models/dense_vnet_abdominal_ct/server.py
```diff
@@ -55,4 +58,4 @@ def seg_file():
 
 
 if __name__ == "__main__":
-    app.run(debug=False, host="127.0.0.1")
+    app.run(debug=True, host="127.0.0.1")
```
nielsboecker, line 61: 👎

HoloPipelines/models/dense_vnet_abdominal_ct/config.ini
```diff
@@ -1,12 +1,10 @@
-############################ input configuration sections
```
nielsboecker, line 1: Why did you delete these comments? They are actually kinda helpful, especially when you start manipulating this file

HoloPipelines/models/dense_vnet_abdominal_ct/server.py
nielsboecker: You can make it debug mode conditionally; take a look how it is handled in the main pipelines server.

HoloPipelines/models/dense_vnet_abdominal_ct/server.py
```diff
@@ -42,7 +43,9 @@ def seg_file():
         if file.filename != UPLOAD_FILENAME:
             file.filename = UPLOAD_FILENAME
         filename = secure_filename(file.filename)
-        file.save(os.path.join(app.config["UPLOAD_FOLDER"], filename))
+        nifty_image_path = os.path.join(app.config["UPLOAD_FOLDER"], filename)
+        file.save(nifty_image_path)
+        config.update_config_with_new_resolution(nifty_image_path)
```
nielsboecker, line 48: Please update README to reflect the two changes you propose (flipping, config)

nielsboecker (commented):
I just skimmed your code, can't run it in Docker at the moment. The code doesn't look bad, but it seems like quite a big change, and a bit late with the deployment already set up. @boonwj will have to greenlight this PR.

My main questions are
* How did you confirm that this all works?
* Do you know which of the two different changes you made led to an improvement?

nielsboecker:
**Update**: #96 implements only one of the two changes proposed here, to reduce the chance of breaking anything.

nielsboecker:
As discussed, this PR changes too many different things and is too risky to merge at this late stage. 
* The feature of changing ini config within the NN models can't be sufficiently tested, so we drop it for now.
* The Y-axis flip for `abdominal_organs_segmentation` and the `split_organs` deprecation node will be integrated with #96

PR #92, opened 2019-08-29
UI: Show live logs and state updates for HoloPipeline jobs

nielsboecker:
This PR will fully integrate the UI with the HoloPipelines:
* Job state is shown and how long this state is already active
* Job log is shown
* After generation is finished, the new hologram shows up in the list

image

> **Disclaimer**: I noticed some design flaws that were made in the UI code early and could only be found now, with the HoloPipelines ready. We can't fix them nicely now due to time constraints, so I had to include some rather hacky work-arounds. I tested it with all three pipelines against the deployed Accessor and Pipelines though, and it always worked. I left notes in the code where it is hacky and should be revisited.


nielsboecker:
**Note to reviewers**: If you want to test against deployed Accessor and HoloPipelines, you can change this in the UI server's `.env` file and then just run UI client and server locally:

```ini
HOLOPIPELINES_HOST=http://51.105.47.56
HOLOPIPELINES_PORT=80
HOLOSTORAGE_ACCESSOR_HOST=http://51.104.216.54
HOLOSTORAGE_ACCESSOR_PORT=80
```

nielsboecker:
There are still some problems with this, due to to Pipelines.
1. When Pipelines job finishes and is already cleaned by garbage collection, it may or may not still be able to send the "FINISHED" state.
1. When Pipelines fails due to invalid input, it won't ever finish by itself, cause Pipelines just fails silently.
1. Even when all is going regular, the Pipelines oten time out at the job state endpoint.

I tried some more fixes for this, but eventually, it is very hard to catch all errors from the UI perspective, especially because different issues have the same symptons. I think it is okay to accept these limitations, and still merge this PR (after code review of course). In worst case, the page needs to be refreshed.

fanbomeng97 (approved):
Have reviewed code and verified that everything works well, good job!

PR #94, opened 2019-08-30
Ui: Colours and README

nielsboecker:
Small change to make the colours look the same (two styling frameworks used different colours so far) and update the main README.


fanbomeng97 (approved):
Nice work！

PR #99, opened 2019-09-04
Misc/doi holo world

boonwj:
Update holo_world steps to include possibility for one to easily change the links in the imaging studies.

### DIFF
--- a/Misc/holo_world/README.md
+++ b/Misc/holo_world/README.md
@@ -7,7 +7,7 @@ This folder contains demo data to setup HoloRepository and the README will docum
 ## Contents
 The following describes the contents provided in the demo.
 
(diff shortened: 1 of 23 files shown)


nielsboecker (approved):
Haha this is insane @boonwj. 🍰 

Just dropping the link in there would have been fine with me, but this is awesome. Maybe the `"address"` field should rather be `"fileName"` now?

boonwj:
Oh the filenames in the address field will be replaced with URLs after running the script.

nielsboecker:
Yeah that makes sense. Maybe the cleanest approach would be a "fileName" field with the filename, and the script then adding a _new_ field with the "address". But let's keep it as is, it's great.

PR #100, opened 2019-09-04
PIPELINE: Update pipeline metadata (less constraints)

nielsboecker:
This is a tiny change. Even though it is in the HoloPipelines dir, it really only affects the UI. This is the JSON that will be sent on the `/pipelines` endpoint. I have updated the "constraints" shown to the user.


boonwj (approved):
Looks okay

Issue #97, opened 2019-08-30
PIPELINE: GLB files have inverted faces after simplification


nielsboecker:
## Description
When in the pipelines a 3D model goes through `simplify`, the obj -> obj conversion seems to go fine and the result looks normal:
![image](https://user-images.githubusercontent.com/11090412/64049210-fbec0900-cb6b-11e9-87b2-d33e2618c447.png)

However, when this OBJ is then transformed to a GLB file by `obj2gltf`, the output looks weird. It seems like the textures outside are transparent, and only on the inside of surfaces, are visible:
![image](https://user-images.githubusercontent.com/11090412/64049594-3ace8e80-cb6d-11e9-8278-8b66b8e31615.png)

## Reproduction with other OBJ file
I reproduced the steps with an OBJ file from the internet (`teddy.obj`) and everything looks fine. Hence, the OBJ files our system produces are the issue.

## Analysis
The problem seems to be that the faces are flipped/inverted. In other words, the norms of these faces point in the exactly wrong direction. Apparently, similar issues [can be caused by coordinate system mismatches](https://forum.babylonjs.com/t/face-ordering-problems-when-replacing-material-on-gltf-imported-mesh/530/3). glTF uses a right-handed coordinate system. When you follow this link and similar Google hits, you will also find that they talk about `invertY` function for 3D loaders as a possible fix.

So, given that the OBJ teddy from the internet works fine, I don't think `simplifier` actually breaks the file. I think the output from the pipeline may already be broken. I find it very suspicious that the guys in the link above talk about `invertY`, just now that you guys flip the Y axis for one of the pipelines (#96).
* Are you 100% sure the problem was there even before you made that flip change?
* Are you completely sure that it really happens for all pipelines, as @ansonwong9695 said?

Furthermore, all files go through the `flip_numpy_array_dimensions` method, which transposes the input and flips one axis. I always found this kind of weird, as you say that it is mirrored, but that would mean that all DICOM files are wrong, or that the default loading method by `pydicom` is broken. Are you really sure we need this??
```python3
def flip_numpy_array_dimensions(array: np.ndarray) -> np.ndarray:
    """
    Transposes numpy axes, i.e. (z, y, x) ---> (x, y, z), and flips x axis. This is
    needed as the default data when we read it is "mirrored" and therefore not accurate
    and, e.g., not valid as input for pre-trained NN models.
    :param array: ndarray from pydicom.read_file()
[...]

PR #101, opened 2019-09-04
PIPELINE: remove the simplify process from pipeline

ansonwong9695:
-Remove the simplify process form each pipeline
-Add comment for why it is unused


nielsboecker (approved):
Okay then, thanks. This sort of fixes the issue I analysed for #97.

boonwj (approved):
Seems like a trivial change.

